PR #85, opened 2026-04-09
fix: enforce on-chain fallback 60s cap for new markets (H2)

0x-SquidSol:
## Summary

- The M-4 on-chain fallback had a 60s time cap to prevent pushing stale prices indefinitely
- The guard `lastExternal > 0` was bypassed for any market that never received an external price (DexScreener/Jupiter), because `lastExternalPriceMs` defaulted to `0`
- A manipulated on-chain `authorityPriceE6` would be pushed forever for new/unlisted markets

## Fix

Seed `lastExternalPriceMs` to `Date.now()` on first encounter of each market:
- New markets get a 60s grace window from their first fallback attempt
- After 60s with no external price, the oracle goes stale naturally (as M-4 intended)
- Remove the `lastExternal > 0` guard since `lastExternal` is now always a real timestamp
- When an external price is eventually fetched, the clock resets normally

## Test plan

- [x] All 133 existing tests pass
- [x] Fix is a 6-line change with clear semantics

### DIFF
--- a/src/services/oracle.ts
+++ b/src/services/oracle.ts
@@ -359,9 +359,15 @@ export class OracleService {
         // failing for longer than ON_CHAIN_FALLBACK_MAX_MS, stop pushing the stale
         // on-chain price. This lets the oracle go stale naturally so the frontend
         // shows a stale warning and the pause guard can halt trading.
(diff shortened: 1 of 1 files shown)


dcccrypto:

PR #89, opened 2026-04-09
fix: remove artificial BigInt overflow guard in liquidation PnL (H7)

0x-SquidSol:
## Summary

- The overflow check clamped PnL at `Number.MAX_SAFE_INTEGER` (9e15), which is meaningless for BigInt (arbitrary precision)
- Large positions where `diff * positionSize > 9e15` were clamped to a **positive** value
- This could make deeply underwater accounts appear healthy, **preventing valid liquidations**
- The re-verification path (line 435) never had this guard and worked correctly

## Fix

Remove the `MAX_SAFE_BIGINT` overflow guard. BigInt arithmetic in JavaScript has arbitrary precision — no overflow is possible. The PnL calculation now matches the re-verification path.

## Test plan

- [x] All 133 tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed large position liquidation PnL calculations by removing unnecessary overflow safeguards that were incorrectly clamping profit/loss values for substantial trading positions, ensuring accurate calculations.

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -256,20 +256,12 @@ export class LiquidationService {
               ? price - entryPrice    // long: profit when price goes up
               : entryPrice - price;   // short: profit when price goes down
             
-            // BH5: Overflow protection - check bounds before multiplication
-            const MAX_SAFE_BIGINT = 9007199254740991n; // Number.MAX_SAFE_INTEGER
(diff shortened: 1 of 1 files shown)


dcccrypto:
Already incorporated: main now uses  (v12.17 change — entryPrice removed from on-chain struct). The BigInt overflow guard block no longer exists in liquidation.ts. Closing as merged-by-evolution.

PR #92, opened 2026-04-09
fix: add timeout to bare getAccountInfo calls in crank service (M1+M2)

0x-SquidSol:
## Summary

Two bare `getAccountInfo` calls in `crank.ts` had no timeout wrapper, allowing hung RPC nodes to block the keeper indefinitely.

**M1 — HYPERP pool lookup (line 503):**
- Fetches DEX pool account to detect type (PumpSwap/Meteora/Raydium)
- A hung RPC blocked the entire crank batch of 10 markets
- Fix: `withTimeout()` + local try/catch — pool data is optional, so timeout degrades gracefully (crank proceeds without vault remaining accounts)

**M2 — MARKETS_FILTER discovery (line 183):**
- Sequential `getAccountInfo` for each filtered slab address at startup
- A hung RPC blocked startup indefinitely
- Fix: `withTimeout()` — existing try/catch catches timeout and continues to next address

## Test plan

- [x] All 133 tests pass
- [x] Both fixes use existing `withTimeout()` and `RPC_TIMEOUT_MS` (15s)

### DIFF
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -180,7 +180,11 @@ export class CrankService {
       for (const addr of slabAddresses) {
         try {
           const pubkey = new PublicKey(addr);
-          const info = await conn.getAccountInfo(pubkey);
+          const info = await withTimeout(
+            conn.getAccountInfo(pubkey),
+            RPC_TIMEOUT_MS,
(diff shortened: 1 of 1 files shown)


dcccrypto:

PR #96, opened 2026-04-09
fix: per-slab single-source counter, per-scan ADL errors (M8+M14)

0x-SquidSol:
## Summary

**M8 — Oracle single-source counter was global:**
- `_consecutiveSingleSource` was a scalar shared across all markets
- Any dual-source market reset the counter, masking degradation for single-source markets
- Now a per-slab `Map` — each market tracks its own cross-validation coverage independently
- Alert includes the affected market slab address for targeted debugging

**M14 — ADL errors counted per-position instead of per-scan:**
- `consecutiveErrors` was incremented for each individual position tx failure inside the loop
- 3 failed positions in one cycle fired the alert meant for 3 consecutive failed *scan cycles*
- Now uses a `scanHadError` flag — counter increments once per failed scan, not per position
- Reset only when a scan has at least one successful send

## Test plan

- [x] All 133 tests pass

### DIFF
--- a/src/services/adl.ts
+++ b/src/services/adl.ts
@@ -438,6 +438,7 @@ export class AdlService {
 
     let sent = 0;
     let remainingExcess = excess;
+    let scanHadError = false;
 
     for (const pos of ranked) {
       if (sent >= ADL_MAX_TX_PER_SCAN) {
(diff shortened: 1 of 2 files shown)


dcccrypto:

PR #101, opened 2026-04-09
fix: remove watchdog force-reset to prevent concurrent cycle overlap (N3)

0x-SquidSol:
## Summary

All three service watchdogs detected hung cycles by resetting the `_cycling`/`_scanning` flag. But the old async operation **continued running** — it was never cancelled. The next interval tick started a **new** cycle, so two cycles ran concurrently:

- **Duplicate transactions** sent to the same markets (wasted SOL)
- **State corruption** from two async operations mutating shared Maps/counters
- The old cycle's `finally` block eventually cleared the flag mid-way through a newer cycle, potentially enabling a **third** concurrent cycle

## Fix

Remove `this._cycling = false` / `this._scanning = false` from all three watchdog branches. The watchdog now **logs + alerts** but does NOT reset the flag. Cycles finish naturally via RPC timeouts (15s per call via `withTimeout`).

**Affected services:**
- `CrankService` — `crank.ts`
- `LiquidationService` — `liquidation.ts`
- `AdlService` — `adl.ts`

## Test plan

- [x] All 133 tests pass
- [x] Updated ADL watchdog test to expect `_cycling = true` (flag preserved) while still verifying alert is sent

### DIFF
--- a/src/services/adl.ts
+++ b/src/services/adl.ts
@@ -554,15 +554,17 @@ export class AdlService {
       if (this._cycling) {
         const elapsed = Date.now() - this._cycleStartedAt;
         if (elapsed > MAX_CYCLE_MS) {
-          logger.error("ADL cycle watchdog: cycle exceeded max duration, force-resetting", {
+          logger.error("ADL cycle watchdog: cycle exceeded max duration — waiting for natural completion", {
             elapsedMs: elapsed,
             maxCycleMs: MAX_CYCLE_MS,
           });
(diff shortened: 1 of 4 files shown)


dcccrypto:
Already incorporated: main already has the N3 fix (never reset _cycling in watchdog), evolved further into H4 with a grace period and process.exit(1) for supervisor restart. The adl.ts change is also incorporated — ADL service is observe-only (no _cycling guard needed). Closing as merged-by-evolution.

PR #105, opened 2026-04-09
fix: wire captureException into critical error handlers (N7)

0x-SquidSol:
## Summary

- `captureException` was imported from `@percolator/shared` but **never called** anywhere in `src/`
- `initSentry("keeper")` initialized Sentry, but zero exceptions were ever reported
- The Sentry integration was effectively dead for the most critical class of errors

## Fix

Add `captureException` calls to three critical error paths:

| Handler | Location | Why |
|---------|----------|-----|
| `unhandledRejection` | `index.ts:567` | Process-level safety net — errors that escaped all try/catch blocks |
| `uncaughtException` | `index.ts:574` | Fatal errors — process is in undefined state |
| `start().catch()` | `index.ts:499` | Startup failure — keeper couldn't start |

Called **before** `logger.error` so Sentry gets the error even if the logger throws.

## Test plan

- [x] All 133 tests pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -497,6 +497,7 @@ async function start() {
 }
 
 start().catch((err) => {
+  captureException(err instanceof Error ? err : new Error(String(err)));
   logger.error("Failed to start keeper", { error: err instanceof Error ? err.message : String(err), stack: err instanceof Error ? err.stack : undefined });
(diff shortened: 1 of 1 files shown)


dcccrypto:

PR #111, opened 2026-05-14
fix(keeper): wire AdlService to notify MonitorService on landed tx

6figpsolseeker:
## Summary

[MonitorService.notifyAdlTx](src/services/monitor.ts#L93) exists and is consumed by the ADL staleness alarm at [src/services/monitor.ts:218-256](src/services/monitor.ts#L218-L256), but a repo-wide grep finds **zero callsites** outside the definition. Because `cycleCountAtLastAdl` stays at its init value of `0` forever, `cyclesSinceLastAdl` equals `_totalCrankCycles` from the moment any market first needs ADL — the staleness alarm fires spuriously on healthy keepers and never resets even when `ExecuteAdl` is working correctly.

This PR wires the missing notification using the existing `CrankService.setOnCrankCycle` pattern:

- [src/services/adl.ts](src/services/adl.ts) — new `_onAdlTx` field + `setOnAdlTx(fn)` setter, mirroring [CrankService at crank.ts:154 and 1156](src/services/crank.ts#L154). Callback fires per landed `ExecuteAdl` tx, right after the existing `sent++` line at [adl.ts:485](src/services/adl.ts#L485).
- [src/index.ts](src/index.ts) — wire `adlService.setOnAdlTx((slab) => monitorService.notifyAdlTx(slab))` alongside the existing `setOnCrankCycle` line at [index.ts:169](src/index.ts#L169). Guarded by the existing `adlService` null check, so deployments with ADL disabled are unaffected.
- [tests/services/adl.test.ts](tests/services/adl.test.ts) — new `describe` block covering per-tx invocation, no-fire on tx failure, and back-compat when no callback is registered.

### Design notes
- **Callback over constructor injection**: matches the established `setOnCrankCycle` precedent and keeps `AdlService`'s zero-arg constructor untouched — no changes needed to the 6 `new AdlService()` callsites in tests.
- **Per-landed-tx (not per-scanMarket)**: `MonitorService._adlTxCounts` increments per call, so per-position notification is more accurate when a multi-tx scan deleverages several positions in one cycle.
- **No `try/catch` around the callback**: matches the bare `this._onCrankCycle?.()` invocation pattern in `CrankService`. `notifyAdlTx` is two `Map` ops on instance state — surfacing real bugs is preferable to swallowing them.

## Test plan

- [x] `pnpm test` — 174 passed, 3 pre-existing skips
- [x] `tsc` compiles clean
- [x] Three new tests in `tests/services/adl.test.ts`:
[...]

PR #110, opened 2026-05-13
fix(keeper): require explicit ADL_ENABLED on mainnet

6figpsolseeker:
## Summary

`ADL_ENABLED` was read with strict equality to `"true"` at [src/index.ts:49](src/index.ts#L49), so an unset or non-matching value silently disabled Auto-Deleveraging. An operator who forgot the var on a mainnet Railway deploy would ship a keeper that never sends `ExecuteAdl`, leaving the protocol exposed when the insurance fund cannot cover losses. The README also marked the var as optional with no mainnet warning.

This PR makes `ADL_ENABLED` a mainnet-required, explicit-only env var:

- New **K-3** rule in [src/env-guards.ts](src/env-guards.ts) — throws on boot when `NETWORK=mainnet` and `ADL_ENABLED` is not exactly `"true"` or `"false"` (after `.trim().toLowerCase()` normalization). Mirrors the existing K-2 SUPABASE pattern and runs before any service constructor.
- [src/index.ts](src/index.ts) — normalize parsing with `.trim().toLowerCase()`; when ADL is deliberately disabled on mainnet (`ADL_ENABLED=false`), fire `sendCriticalAlert` so the off state pages ops loudly instead of staying silent.
- Devnet and testnet keep the legacy permissive default, so existing local and CI workflows are unaffected.
- [README.md](README.md) and [.env.example](.env.example) updated to document the new requirement.

### Why not flip the default to `true` on mainnet
`AdlService.scanMarket` calls `encodeExecuteAdl` and `sendWithRetryKeeper` unconditionally — there is no on-chain feature-detect for the T8 instruction (PERC-8273) or the T10 devnet upgrade (PERC-8275). A silent default-on would make any operator pulling main start sending failed `ExecuteAdl` txs on chains where T8 has not landed, burning SOL and tripping the consecutive-failure watchdog. An explicit-required guard surfaces the decision without changing program semantics.

## Test plan

- [x] `pnpm test` passes (178 passed, 3 pre-existing skips)
- [x] `tests/env-guards.test.ts` covers: unset on mainnet (throws), invalid value on mainnet (throws), `"true"` on mainnet (passes), `"false"` on mainnet (passes), case/whitespace normalization (`"  True  "` passes), devnet permissive (no var → no throw), `NETWORK` unset (no throw)
- [x] `tsc` compiles clean
- [ ] Manual: deploy with `NETWORK=mainnet` and unset `ADL_ENABLED` — keeper must refuse to start with the K-3 error
[...]

PR #112, opened 2026-05-15
fix(keeper): abort liquidation on oracle drift between scan and submit

6figpsolseeker:
## Summary

`encodeLiquidateAtOracle` is a fixed 3-byte instruction (`tag + u16 targetIdx` — verified in `node_modules/@percolatorct/sdk/dist/abi/instructions.d.ts:349-352` and `dist/index.js:403-408`) and cannot carry an oracle price bound. If the oracle moves between when [scanMarket](src/services/liquidation.ts) decides a candidate is liquidatable and when [liquidate](src/services/liquidation.ts) sends the tx, the on-chain execution price differs from the keeper's assumption and the keeper-controlled wallet absorbs the difference.

This PR adds a keeper-side drift guard:

1. [src/services/liquidation.ts](src/services/liquidation.ts) — new `MAX_LIQUIDATION_DRIFT_BPS` constant (default 150 bps, configurable via `LIQUIDATION_MAX_ORACLE_DRIFT_BPS`; set to `0` to disable).
2. New `scanPriceE6: bigint` field on `LiquidationCandidate` capturing the price resolved by `resolveMarketPrice` at candidacy time.
3. Inside the existing pre-check `freshData` block — right where `freshPrice` is already computed — compare against `scanPriceE6` and abort with a `logger.warn` if drift exceeds the limit.
4. `scanAndLiquidateAll` callsite updated to thread `candidate.scanPriceE6` into `liquidate()`. The third parameter defaults to `0n` so any existing direct callers (e.g. tests) keep working with the guard skipped.

### Window covered
This bounds keeper exposure to oracle drift in the **scan→pre-check** window — the dominant gap, up to `intervalMs = 60_000` by default plus per-batch overhead. The narrower pre-check→submit window (~50-200ms) is **not** covered and can be added later if production metrics show meaningful drift there.

### Why a keeper-side mitigation rather than a program-side bound
The on-chain `LiquidateAtOracle` instruction wire format is 3 bytes — there is no field to populate. A true on-chain bound requires a program + SDK change. This PR is the highest-impact protection available from the keeper alone.

### Why 150 bps default
~2-3 sigma of typical intra-minute moves on SOL/BTC/ETH; wide enough to avoid spurious aborts during normal volatility, tight enough to cap exposure on a 60s scan interval. Tune via `LIQUIDATION_MAX_ORACLE_DRIFT_BPS` if mainnet observations suggest otherwise.

## Test plan

- [x] `pnpm test` — 174 passed, 3 pre-existing skips
- [x] `tsc` compiles clean
- [x] Three new tests in `tests/services/liquidation.test.ts`:
  - aborts when drift exceeds 150 bps (returns null, `sendWithRetryKeeper` not called)
  - proceeds when drift is within limit (sig returned, `sendWithRetryKeeper` called)
  - back-compat: skips the guard when `scanPriceE6` defaults to `0n`
- [ ] Manual on devnet: observe `liquidation.scanCount` and the new warn-log to size actual drift frequency before/after.

## Notes

Composes cleanly with #110 (`ADL_ENABLED` mainnet guard) and #111 (wire ADL→Monitor) — independent code paths, no overlap.

A follow-up could add: (a) a second slab fetch right before submit to bound the pre-check→submit window, or (b) a metrics counter on aborts so the team can size the residual risk.

## Release Notes

* **New Features**
  * Liquidation execution now includes an oracle-drift guard. The system records oracle prices during the scanning phase and compares them against prices at execution time. If prices differ beyond the configured threshold (default: 150 basis points), liquidation is aborted. The drift tolerance is configurable via environment variables.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dcccrypto/percolator-keeper/pull/112)

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -78,6 +78,18 @@ async function fetchSlabWithRetry(
 const PRICE_E6_DIVISOR = 1_000_000n; // Price precision divisor (6 decimals)
 const BPS_MULTIPLIER = 10_000n; // Basis points multiplier (100% = 10000 bps)
 
+// Oracle-drift guard: abort liquidation if the oracle price has drifted more
+// than this between scan-time candidacy and pre-submit re-verification.
+// The on-chain LiquidateAtOracle instruction is 3 bytes (tag + u16 targetIdx —
+// see @percolatorct/sdk encodeLiquidateAtOracle) and cannot carry a [min,max]
+// bound, so keeper-side drift detection is the only available mitigation.
+// Default 150 bps (1.5%) — wider than typical intra-minute moves on SOL/BTC/ETH
+// but tight enough to cap keeper-wallet exposure on a 60s scan interval.
+// Set to 0 to disable the guard.
+const MAX_LIQUIDATION_DRIFT_BPS = BigInt(
+  parseInt(process.env.LIQUIDATION_MAX_ORACLE_DRIFT_BPS ?? "150", 10),
+);
+
 /**
  * Oracle mode for a market.
  * - 'pyth-pinned': oracle_authority == [0;32] && index_feed_id != [0;32]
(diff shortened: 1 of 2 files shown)


dcccrypto:
Superseded by #204 (rebased on current main, re-expressed for the current liquidation.ts structure — no LiquidateAtOracle instruction in current main, guard applies to the Liquidate instruction instead).

PR #25, opened 2026-03-27
feat(PERC-8108): mainnet HYPERP oracle wiring + stale mark pause guard

dcccrypto:
## Summary

Wires keeper to use HYPERP oracle for mainnet markets (SOL-PERP, BTC-PERP, ETH-PERP) and adds stale mark pause guard.

### New files
- `src/config/network.ts` — `isMainnet()` helper, reads `NETWORK` env var
- `src/config/mainnet-markets.ts` — mainnet HYPERP market mint constants (SOL, BTC, ETH)
- `tests/services/oracle-stale.test.ts` — 5 tests for staleness detection logic

### Modified files
- `src/services/oracle.ts` — `getStaleMarkets()` method: returns markets not updated in >N minutes
- `src/services/crank.ts` — wires stale pause callback from CrankService
- `src/index.ts` — NETWORK=mainnet switches HYPERP config; stale alert at 5min, pause crank at 10min; /pause-status health endpoint
- `.env.example` — documents `NETWORK`, `MAINNET_RPC_URL`, `MAINNET_PROGRAM_ID`, `MAINNET_SUPABASE_URL`, `MAINNET_SUPABASE_SERVICE_KEY`

## Behaviour
- **devnet (default):** no change
- **NETWORK=mainnet:** keeper uses mainnet RPC + program ID + Supabase; HYPERP markets active
- **Stale oracle guard:** if oracle not updated in >5min → WARN; >10min → pause mark updates + alert

## Testing
- 109 tests pass (6 suites), including 5 new oracle-stale tests

## Notes
- No mainnet deploy — code + config ready for devops to deploy
- Related: PERC-8108

Closes PERC-8108

## Summary by CodeRabbit

* **New Features**
  * Added mainnet mode support with configurable environment variables
  * Implemented stale oracle detection with automatic market pausing (5-min alert, 10-min pause thresholds)
  * Added critical alert notifications when oracle data becomes stale
  * New `/pause-status` endpoint to monitor paused markets

* **Tests**
  * Added comprehensive tests for stale oracle detection functionality

PR #120, opened 2026-05-26
feat(keeper): Phase 1.5 Workstream G — Alchemy secondary RPC + read/write split

dcccrypto:
## What this adds

Three new keeper-internal modules providing RPC provider redundancy:

- **`src/lib/rpc-health.ts`** — per-provider P50/P99 latency tracker. Rolling 60-second window of latency samples (sorted-sample nearest-rank percentile), slot-lag tracking, and a three-criterion recovery gate. A provider only recovers if: P99 < 1000ms AND |slot lag| < 10 AND 0 consecutive fails, sustained for the full `recoveryWindowMs` (60s default). Any criterion failure resets the recovery clock.
- **`src/config/rpc.ts`** — env-var config parser. `ALCHEMY_RPC_URL` is validated only when `RPC_POOL_ENABLED=true` so existing keeper deployments without Alchemy configured boot without error.
- **`src/lib/rpc-pool.ts`** — generic pool wrapping two `Connection` instances. Reads route to Helius while healthy; fall back to Alchemy on Helius unhealthy; fail-safe returns Helius when both are unhealthy (degraded primary > no primary). `getProgramAccounts` prefers Alchemy when `RPC_FORCE_ALCHEMY_GPA=true` (default), but falls back to Helius if Alchemy is unhealthy. `writeConnection` getter always returns Helius unconditionally.

Extends **`src/lib/metrics.ts`** with five new exported metrics:
- `keeper_rpc_request_total{provider, method, result}`
- `keeper_rpc_latency_ms{provider, percentile}` (shared gauge, p50/p99 label)
- `keeper_rpc_provider_healthy{provider}`
- `keeper_rpc_failover_total{from, to, reason}`
- `keeper_rpc_slot_lag{provider}`

## Why

Provider redundancy without touching writes. Helius failure currently means all keeper reads fail. Alchemy as a health-checked failover gives reads a second path. Writes, LaserStream (Helius WebSocket), and `keeperSend` are structurally untouched.

## What this does NOT do

- No migration of existing `getConnection()` call sites — Workstream G is additive only. A follow-up PR will wire callers.
- No write-path changes. `keeperSend` and `keeperSend`-adjacent code are untouched.
- No LaserStream changes (`src/lib/account-loader.ts`, `src/lib/account-cache.ts` unchanged).
- No new npm dependencies. Uses Node 22 built-in fetch, existing `@solana/web3.js` `Connection`, `prom-client`.

## Test plan

| File | Type | Count |
|---|---|---|
| `tests/lib/rpc-health.test.ts` | Unit | 25 tests |
| `tests/lib/rpc-pool.test.ts` | Unit | 28 tests |
| `tests/lib/rpc-pool.property.test.ts` | Property (fast-check, 500 runs each) | 5 invariants |
| `tests/lib/rpc-pool.stress.test.ts` | Stress/chaos (STRESS=true gated) | 4 scenarios |

**Non-STRESS run:** 501 passing / 20 skipped
**STRESS=true run:** 516 passing / 5 skipped
**pnpm build:** clean (0 TypeScript errors, strict mode)

STRESS scenarios verified:
- 1000 concurrent `read()` during simulated provider failure — all 1000 complete, zero drops
- Chaos: Helius 500s — all reads route to Alchemy within one health-check cycle
- Chaos: both providers unhealthy simultaneously — fail-safe routes to Helius, no exception thrown
- Chaos: Helius recovers after failure — reads return to Helius after 60s clean window

## Operator config (new env vars)

Add to Railway/`.env`:

```bash
# Required when RPC_POOL_ENABLED=true:
ALCHEMY_RPC_URL=https://solana-mainnet.g.alchemy.com/v2/YOUR_KEY

# Optional tuning (shown with defaults):
RPC_POOL_ENABLED=false          # false = always Helius (safe default; set true to enable pool)
RPC_HEALTH_CHECK_INTERVAL_MS=5000
RPC_UNHEALTHY_P99_MS=2000
RPC_UNHEALTHY_SLOT_LAG=50
RPC_UNHEALTHY_CONSECUTIVE_FAILS=5
RPC_RECOVERY_WINDOW_MS=60000
RPC_FORCE_ALCHEMY_GPA=true
```

`RPC_POOL_ENABLED` defaults to `false`. Set to `true` once `ALCHEMY_RPC_URL` is configured.

## Follow-up

A separate PR will migrate existing `getConnection()` callers in services to use `sharedRpcPool().getAccountInfo(...)` etc. for reads. This PR intentionally ships the pool without touching call sites.

## Non-blocking follow-up: Workstream G.1 — slot sanity bound

Security item #25 (not in scope of original brief): Reject Alchemy `getSlot` responses where `slot > heliusSlot + MAX_REASONABLE_LAG (~1000)` to prevent slot-manipulation route hijacking. Tracked for a separate hardening PR.

## Summary by CodeRabbit

* **New Features**
  * Added RPC pool failover mechanism that automatically routes requests to a secondary provider when the primary provider becomes unhealthy.
  * Implemented health monitoring with configurable thresholds for latency, slot lag, and consecutive failures to detect provider issues.
  * Added metrics for tracking RPC request routing, provider health status, and failover events.

* **Tests**
  * Added comprehensive test coverage for RPC pool failover behavior and health monitoring.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dcccrypto/percolator-keeper/pull/120?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -81,3 +81,21 @@ USE_HELIUS_SENDER=true
 # KEEPER_LEADER_LOCK_TTL_MS=30000
 # KEEPER_LEADER_LOCK_RENEW_MS=10000
 # KEEPER_STANDBY_POLL_MS=5000
+
+# RPC pool (Alchemy secondary) — Workstream G
+# Helius stays primary for reads and the sole write path. Alchemy is health-checked
+# failover: reads fall back to Alchemy when Helius P99 > 2s, slot lag > 50, or
+# 5+ consecutive health-check failures. LaserStream (WebSocket) stays on Helius.
+# getProgramAccounts always routes to Alchemy by default (heavy method, spares Helius).
+#
+# Required when RPC_POOL_ENABLED=true:
+# ALCHEMY_RPC_URL=https://solana-mainnet.g.alchemy.com/v2/YOUR_KEY
+# ALCHEMY_API_KEY=                    # informational; the key is embedded in ALCHEMY_RPC_URL
+#
+# RPC_POOL_ENABLED=true               # false = always Helius (current behaviour pre-G)
+# RPC_HEALTH_CHECK_INTERVAL_MS=5000   # how often to probe both providers
+# RPC_UNHEALTHY_P99_MS=2000           # P99 > this → Helius unhealthy
+# RPC_UNHEALTHY_SLOT_LAG=50           # |slot lag| > this → unhealthy
+# RPC_UNHEALTHY_CONSECUTIVE_FAILS=5   # consecutive probe failures → unhealthy
+# RPC_RECOVERY_WINDOW_MS=60000        # must hold clean for this long before re-promoting Helius
+# RPC_FORCE_ALCHEMY_GPA=true          # always route getProgramAccounts to Alchemy
--- a/src/config/rpc.ts
+++ b/src/config/rpc.ts
@@ -1,0 +1,75 @@
+/**
+ * RPC provider configuration and routing rules for the keeper RPC pool.
+ *
+ * Parsed once at startup. ALCHEMY_RPC_URL and ALCHEMY_API_KEY are validated
+ * only when RPC_POOL_ENABLED=true — this avoids breaking keeper boots that
+ * have not yet configured Alchemy.
+ */
+
+function parseBoolEnv(env: NodeJS.ProcessEnv, key: string, defaultVal: boolean): boolean {
+  const raw = env[key];
+  if (raw === undefined || raw === "") return defaultVal;
+  return raw.trim().toLowerCase() === "true";
+}
+
+function parseIntEnv(env: NodeJS.ProcessEnv, key: string, defaultVal: number): number {
+  const raw = env[key];
+  if (raw === undefined || raw === "") return defaultVal;
+  const n = Number(raw);
+  if (!Number.isFinite(n) || n <= 0 || !Number.isInteger(n)) return defaultVal;
+  return n;
+}
+
+export interface RpcProviderConfig {
+  url: string;
+  name: "helius" | "alchemy";
+}
+
+export interface RpcPoolConfig {
+  enabled: boolean;
+  helius: RpcProviderConfig;
+  alchemy: RpcProviderConfig;
+  healthCheckIntervalMs: number;
+  unhealthyP99Ms: number;
+  unhealthySlotLag: number;
+  unhealthyConsecutiveFails: number;
+  recoveryWindowMs: number;
+  forceAlchemyGpa: boolean;
+}
+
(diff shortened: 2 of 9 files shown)


dcccrypto:
## Review feedback addressed — commit 35dd7d6

All 3 MED blockers, 1 FAIL, and 1 spec deviation fixed:

**Fix 1 — `RPC_POOL_ENABLED` default → `false`** (`src/config/rpc.ts:41`): Changed from `true` to `false`; pool is now opt-in. Existing deployments without `ALCHEMY_RPC_URL` boot cleanly.

**Fix 2 — Null `otherLastSeenSlot` blocks recovery** (`src/lib/rpc-health.ts:149`): Changed `slotLag === null ||` to `slotLag !== null &&`. Recovery now requires affirmative slot data from the other provider. New unit test verifies recovery stays blocked through a 120s advance when other provider slot is null, then confirms recovery proceeds once slot data arrives.

**Fix 3 — GPA falls back to Helius when Alchemy unhealthy** (`src/lib/rpc-pool.ts:pickProvider`): `forceAlchemyGpa=true` + Alchemy unhealthy now falls back to Helius instead of returning unhealthy Alchemy. Bumps `keeper_rpc_failover_total{from:alchemy, to:helius, reason:gpa_alchemy_unhealthy}`. Three new unit tests cover healthy path, unhealthy fallback, and counter assertion.

**Fix 4 — Property test no longer vacuous** (`tests/lib/rpc-pool.property.test.ts`): Replaced `applyEvents` path with `pool.tickForTest()`, a new synchronous method that injects scripted (heliusSlot, alchemySlot, latencyMs) tuples and runs `_evaluateAndTransition()` — the real production state machine. `_healthTick` was refactored to call `_evaluateAndTransition` internally, eliminating duplication. 500-run property now catches any decrement of `_failoverCount`.

**Fix 5 — Discord alert on failover** (`src/lib/rpc-pool.ts:_evaluateAndTransition`): `sendWarningAlert("RPC failover", [{From}, {To}, {Reason}])?.catch(() => {})` fires on both failover directions. No URL or API key in payload.

**Fix 6 — Path-aware URL redaction** (`src/lib/rpc-pool.ts:start`): Replaced `.slice(0, 40)` with `.replace(/\/v2\/.*/i, '/v2/<redacted>')`. Three new unit tests verify redaction works for standard Alchemy URL shape, short custom URL shape, and no-op on URLs without `/v2/`.

**Test counts after fixes:** 501 passing (non-STRESS) / 516 passing (STRESS=true) / 0 failures.

---

**Non-blocking follow-up (Workstream G.1):** Slot sanity bound — reject Alchemy `getSlot` responses where `slot > heliusSlot + 1000` to prevent slot-manipulation route hijacking. Not in scope of original brief; tracked for a future hardening PR.

PR #131, opened 2026-06-04
fix(liquidation): H3 — use Solana Clock sysvar for admin-oracle staleness

6figpsolseeker:
## Summary

Fixes **H3** from the internal pre-mainnet audit. Replaces the wall-clock (`Date.now()/1000`) timestamp comparison in `resolveMarketPrice` with the Solana **Clock sysvar** unix_timestamp, so the keeper's admin-oracle staleness check matches the on-chain program's check byte-for-byte.

## Root cause

```ts
// before: wall clock
const now = BigInt(Math.floor(Date.now() / 1000));
const priceAge = cfg.authorityTimestamp > 0n ? now - cfg.authorityTimestamp : now;
const authorityFresh = cfg.authorityPriceE6 > 0n && priceAge <= 60n;
```

`cfg.authorityTimestamp` is written **on-chain** using the Solana Clock sysvar (cluster time). Wall clock and cluster time can drift by seconds during NTP step corrections, container clock skew, or validator congestion.

## Threat model

The keeper already passes `SYSVAR_CLOCK_PUBKEY` into both `keeper_crank` and `liquidate_at_oracle` instructions, so the on-chain handler does its own freshness check using the Clock sysvar. The keeper's check is therefore a **pre-filter** to avoid wasted gas, not the security boundary:

- The 60s window absorbs realistic NTP-synced drift (sub-second to ~10s in the worst documented cases).
- Worst-case impact: a wall-clock-skewed keeper submits a tx the on-chain instruction rejects (wasted gas / failed tx), or falls back to `lastEffectivePriceE6` when the authority price is actually fresh on-chain.
- **No wrongful liquidation** — the on-chain program is the source of truth for collateralization decisions.

**Severity: LOW** (downgraded from initial HIGH). Correctness hygiene / audit cleanliness.

## Fix

1. Add `fetchClusterUnixTimeSec()` helper:
   ```ts
   async function fetchClusterUnixTimeSec(): Promise<bigint> {
     try {
       const info = await getConnection().getAccountInfo(SYSVAR_CLOCK_PUBKEY);
       if (info && info.data && info.data.length >= 40) {
         return info.data.readBigInt64LE(32); // unix_timestamp field
       }
       logger.warn("H3: SYSVAR_CLOCK_PUBKEY account info malformed, falling back to wall clock", ...);
     } catch (err) {
       logger.warn("H3: failed to fetch Clock sysvar, falling back to wall clock", ...);
     }
     return BigInt(Math.floor(Date.now() / 1000));
   }
   ```
2. `resolveMarketPrice` now takes `nowSec: bigint` as a parameter.
3. Both call sites (scanMarket and the pre-submit recheck) fetch cluster time once and pass it in.

**Clock sysvar layout** (40 bytes total, all little-endian):

| offset | field | type |
|---|---|---|
| 0  | slot | u64 |
| 8  | epoch_start_timestamp | i64 |
| 16 | epoch | u64 |
| 24 | leader_schedule_epoch | u64 |
| 32 | **unix_timestamp** | **i64** ← what we read |

## Files touched

- `src/services/liquidation.ts` — add helper, modify `resolveMarketPrice` signature, wire cluster time at both call sites.
- `tests/services/liquidation.test.ts` — 4 new tests.

## Test plan

- [x] **Clock sysvar fetched during liquidate()**: proves the cluster-time path executes (asserts SYSVAR_CLOCK_PUBKEY appears in `getAccountInfo` call list).
- [x] **Falls back to wall clock on RPC failure**: connection throws → keeper still completes via `Date.now()` fallback.
- [x] **Falls back to wall clock on malformed Clock data**: returns 8-byte buffer (too short) → keeper falls back via `Date.now()`.
- [x] **scanMarket also fetches Clock sysvar**: both call sites use cluster time, not just `liquidate()`.
- [x] Liquidation suite: **17 passed** (was 13).
- [x] Full vitest suite: **616 passed / 26 skipped** — no regressions.
- [x] `pnpm build` clean.

## Risk

- **One extra RPC call per `liquidate()` and per `scanMarket()` invocation.** Negligible vs. `fetchSlab` already done per market; no new throughput pressure.
- **Backward-compatible at the deployment layer.** No new config required.
- **Robust against RPC outages** — automatic fallback to wall clock preserves keeper liveness.

## Summary by CodeRabbit

* **Bug Fixes**
  * Use Solana cluster clock for price staleness checks (with fallback to local time) to align behavior with on-chain validation.

* **New Features**
  * Added a utility to compute margin ratio for clearer liquidation decisions.

* **Tests**
  * Added unit, integration, smoke and PoC tests validating cluster-time handling, fallback behavior, and simulator Clock round-trips.

* **Chores**
  * Added a dev dependency to support simulator tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "devDependencies": {
     "@types/node": "^25.2.2",
     "fast-check": "^4.8.0",
+    "litesvm": "^1.1.0",
     "tsx": "^4.19.2",
     "typescript": "^5.7.3",
     "vitest": "^4.0.18"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -45,6 +45,9 @@ importers:
       fast-check:
         specifier: ^4.8.0
         version: 4.8.0
+      litesvm:
+        specifier: ^1.1.0
+        version: 1.1.0(bufferutil@4.1.0)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(utf-8-validate@6.0.6)
       tsx:
         specifier: ^4.19.2
         version: 4.21.0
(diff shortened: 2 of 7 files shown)


dcccrypto:
Closing in favor of clean rebase against current main. Fix has been applied and a new PR created.

PR #134, opened 2026-06-08
fix(keeper): make ADL observe-only to align with admin-gated ExecuteAdl

### LINKED ISSUE #133: Keeper ADL service cannot execute ExecuteAdl against current program (admin-gated + insurance-depletion preconditions)
## Summary

The keeper's ADL service (`src/services/adl.ts`) assumes `ExecuteAdl` (instruction tag 50) is a permissionless keeper operation. On the current program this is no longer true: `ExecuteAdl` is **admin-gated** and additionally requires the market's insurance fund to be **fully depleted**. As written, every `ExecuteAdl` the keeper can send is rejected on-chain.

This is latent today because the service is gated behind `ADL_ENABLED`, which defaults to off. It becomes a 100%-failure path the moment ADL is enabled with a normal (non-admin) crank key.

## Program behavior (current `main`)

`handle_execute_adl` enforces, in order:

- `require_admin(header.admin, accounts[0])` — the signer must be the market admin, otherwise `EngineUnauthorized`.
- `if insurance_fund.balance != 0 { return InsuranceFundNotDepleted }` — ADL is only admissible once the insurance fund is fully depleted.
- if `max_pnl_cap > 0`, then `pnl_pos_tot` must exceed it.

(Reference: `dcccrypto/percolator-prog`, `src/percolator.rs`, `handle_execute_adl`.)

## Keeper behavior

- The ADL service signs `ExecuteAdl` with the operational crank key (`CRANK_KEYPAIR`), which is not the market admin → fails `require_admin`.
- The trigger fires on `pnl_pos_tot > max_pnl_cap` or an insurance-utilization ratio (default ~80%), neither of which requires `balance == 0` → fails `InsuranceFundNotDepleted` even if the signer were the admin.

Net effect once enabled: a stream of guaranteed-reverting transactions that burn fees and inflate failure/alert counters, with no deleveraging actually performed.

## Why this is safe to change

Routine, protective deleveraging already happens permissionlessly inside `KeeperCrank` (which runs the engine's progress/recovery path) and `LiquidateAtOracle`, both of which the keeper already sends. `ExecuteAdl` is a targeted admin/multisig action with no on-chain ranking verification. Because the keeper's `ExecuteAdl` sends always revert, removing that path loses no protection.

## Supporting cleanups

- `src/services/crank-types.ts` carries unused two-phase `PrepareAdl` scaffolding (including a `0xff` placeholder tag) for an on-chain prepare/execute split that does not exist — the program's ADL is single-phase.
- `src/services/monitor.ts` raises an "ADL needed" alert based on `pnl_pos_tot > max_pnl_cap` alone, producing false positives for healthy markets whose insurance fund still holds a balance.

## Reproduction (PoC)

The following test uses the genuine `@percolatorct/sdk` encoders (the same functions the keeper calls) plus the keeper's own trigger logic. It demonstrates both gate violations against current `main` without requiring a validator:

```ts
import { describe, it, expect, vi } from "vitest";
import { Keypair, PublicKey, SYSVAR_CLOCK_PUBKEY } from "@solana/web3.js";

// Real SDK encoders; only the on-chain readers are stubbed.
vi.mock("@percolatorct/sdk", async () => {
  const actual = await vi.importActual<typeof import("@percolatorct/sdk")>("@percolatorct/sdk");
  return { ...actual, fetchSlab: vi.fn(), parseEngine: vi.fn(), parseConfig: vi.fn(), parseAllAccounts: vi.fn(() => []) };
});
  getConnection: vi.fn(() => ({})),
  loadKeypair: vi.fn(() => Keypair.generate()), // crank key is an ordinary hot wallet, not the admin
  createLogger: vi.fn(() => ({ info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() })),
  sendWarningAlert: vi.fn(async () => {}),
  sendCriticalAlert: vi.fn(async () => {}),
}));

import * as sdk from "@percolatorct/sdk";
import { encodeExecuteAdl, buildAccountMetas, ACCOUNTS_EXECUTE_ADL } from "@percolatorct/sdk";
import { AdlService } from "../../src/services/adl.js";

describe("PoC: keeper ExecuteAdl violates the program's gates", () => {
  it("signs ExecuteAdl with the crank key, not the market admin (require_admin rejects)", () => {
    const crankKey = Keypair.generate().publicKey;
    const marketAdmin = Keypair.generate().publicKey;

    // Built exactly as the keeper builds it, using the genuine SDK encoders.
    const data = encodeExecuteAdl({ targetIdx: 3 });
    const keys = buildAccountMetas(ACCOUNTS_EXECUTE_ADL, [
      crankKey,
      Keypair.generate().publicKey, // slab
      SYSVAR_CLOCK_PUBKEY,
      Keypair.generate().publicKey, // oracle
    ]);

    expect(data[0]).toBe(50); // tag 50, encoding is correct
    expect(keys[0].isSigner).toBe(true);
    expect((keys[0].pubkey as PublicKey).equals(crankKey)).toBe(true);
    expect(crankKey.equals(marketAdmin)).toBe(false); // → require_admin(header.admin, accounts[0]) rejects
  });

  it("triggers ADL while insurance_fund.balance > 0 (InsuranceFundNotDepleted rejects)", async () => {
    vi.mocked(sdk.fetchSlab).mockResolvedValue(new Uint8Array(1024));
    vi.mocked(sdk.parseEngine).mockReturnValue({
      pnlPosTot: 1_000_000n,
      insuranceFund: { balance: 5_000_000n, feeRevenue: 0n },
    } as any);
    vi.mocked(sdk.parseConfig).mockReturnValue({ maxPnlCap: 500_000n } as any);

    const svc = new AdlService();
    const state = await svc.getAdlState("slab", { slabAddress: Keypair.generate().publicKey } as any);

    expect(state.adlNeeded).toBe(true); // the keeper will send ExecuteAdl …
    expect(BigInt(state.insuranceFundBalance)).toBeGreaterThan(0n); // … but the fund is not depleted → reverts
  });
});
```

## Proposed fix

Make the keeper's ADL service **observe-only**: remove the sign/send path, keep the read-only state/rankings for `/api/adl/rankings` and a rate-limited "admin/multisig action required" alert, and gate the reported condition on `insurance_fund.balance == 0 && (max_pnl_cap == 0 ? pnl_pos_tot > 0 : pnl_pos_tot > max_pnl_cap)`. Targeted `ExecuteAdl` remains an admin/multisig runbook action; protective deleveraging stays with the permissionless crank/liquidate paths.

A PR implementing this (with a regression test) follows.

0x-SquidSol:
## Summary

Fixes #133. Makes the keeper's ADL service **observe-only** to align with the current program, where `ExecuteAdl` (tag 50) is admin-gated and requires the insurance fund to be fully depleted.

## Background

`handle_execute_adl` enforces `require_admin(header.admin, accounts[0])` and rejects unless `insurance_fund.balance == 0`. The keeper signed `ExecuteAdl` with the operational crank key (not the admin) and triggered on PnL-cap / insurance-utilization without requiring `balance == 0` — so every send would revert (`EngineUnauthorized` / `InsuranceFundNotDepleted`). Protective deleveraging already happens permissionlessly inside `KeeperCrank` and `LiquidateAtOracle` (both already sent by the keeper), so targeted `ExecuteAdl` is an admin/multisig action. Because the keeper's `ExecuteAdl` sends always reverted, removing the path loses no protection.

## Changes

- **`src/services/adl.ts`** — remove the sign/send path and the crank-key signer. `AdlService` is now observe-only: it evaluates the on-chain ADL preconditions per market and exposes them via `getAdlState()`/`getStats()` (for `/api/adl/rankings`) plus a rate-limited "admin/multisig action required" alert. `adlNeeded` now mirrors the program: `balance == 0 && (cap == 0 ? pnl > 0 : pnl > cap)`. Utilization is informational only.
- **`src/services/monitor.ts`** — gate the ADL-staleness alert on `insurance_fund.balance == 0` and reword it to "admin/multisig action required" (the keeper no longer executes ADL).
- **`src/services/crank-types.ts`** — remove the unused two-phase `PrepareAdl` scaffolding (incl. the `0xff` placeholder tag) for an on-chain split that does not exist; the program's ADL is single-phase.
- **`src/index.ts`** — report ADL status as observe-only.
- **`tests/`** — rewrite the ADL suite for observe-only behavior and add a regression test asserting that (a) `adlNeeded` is false while `balance > 0`, and (b) no `ExecuteAdl` is ever encoded/built/sent even when preconditions are met.

## Verification

- `tsc --noEmit`: clean (exit 0).
- Full test suite: **609 passed, 26 skipped, 0 failed**.
- The regression spies on `encodeExecuteAdl` / `buildIx` / `buildAccountMetas` / `sendWithRetryKeeper` will fail if a keeper-side ADL send path is ever reintroduced.

## Notes

- No behavior change in production today — the service is gated behind `ADL_ENABLED`, which defaults to off.
- No protocol-safety loss: the removed sends always reverted; permissionless crank/liquidate paths are unchanged.

## Summary by CodeRabbit

* **New Features**
  * ADL now operates in observe-only mode, providing enhanced market monitoring and automated alerts when critical conditions are detected.
  * Health endpoint updated to report ADL observe-only status and counts of markets requiring attention.

* **Bug Fixes**
  * Improved ADL trigger condition evaluation and updated alert behavior to accurately reflect system state.

* **Tests**
  * Added regression test suite validating observe-only ADL behavior and preventing unintended transaction execution.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,14 +57,17 @@ const liquidationService = new LiquidationService(oracleService);
 const monitorService = new MonitorService();
 const fraudDetector = new FraudDetectorService(oracleService, () => crankService.getMarkets());
 
-// ADL service — gated by ADL_ENABLED=true env var until on-chain instruction
-// (PERC-8273 T8) is live and T10 devnet upgrade is done (PERC-8275).
+// ADL service — OBSERVE-ONLY, gated by ADL_ENABLED=true.
+// ExecuteAdl is admin/multisig-gated on-chain (require_admin + insurance must be
+// fully depleted); the keeper does NOT send it. Protective deleveraging happens
+// permissionlessly inside KeeperCrank/LiquidateAtOracle. This service only reports
+// when on-chain ADL preconditions are met. See services/adl.ts header.
 const adlEnabled = process.env.ADL_ENABLED === "true";
 const adlService = adlEnabled ? new AdlService() : null;
 if (adlEnabled) {
-  logger.info("ADL service enabled (ADL_ENABLED=true)");
+  logger.info("ADL service enabled (observe-only — ExecuteAdl is admin/multisig-gated)");
 } else {
-  logger.info("ADL service disabled — set ADL_ENABLED=true to enable (requires T8+T10)");
+  logger.info("ADL service disabled — set ADL_ENABLED=true for observe-only ADL monitoring");
 }
 
 // HA leader lock — null when HA_ENABLED is not set or KEEPER_REDIS_URL is absent
@@ -453,17 +456,15 @@ res.writeHead(401, secureJsonHeaders);
       }
     }
     
-    // ADL stats
+    // ADL stats — observe-only (ExecuteAdl is admin/multisig-gated; keeper does not send)
     let adlStats: Record<string, unknown> | null = null;
     if (adlService) {
       const stats = adlService.getStats();
-      let totalAdlTxSent = 0;
-      let activeMarkets = 0;
+      let marketsNeedingAdl = 0;
       for (const [, s] of stats) {
-        totalAdlTxSent += s.adlTxSent;
-        if (s.adlTxSent > 0) activeMarkets++;
+        if (s.adlNeeded) marketsNeedingAdl++;
       }
-      adlStats = { enabled: true, totalAdlTxSent, activeMarkets };
+      adlStats = { enabled: true, mode: "observe-only", marketsNeedingAdl };
     } else {
       adlStats = { enabled: false };
     }
@@ -649,7 +650,7 @@ async function start() {
 
     if (adlService) {
       adlService.start(() => crankService.getMarkets());
-      logger.info("ADL service started");
+      logger.info("ADL service started (observe-only)");
     }
   }
 
--- a/src/services/adl.ts
+++ b/src/services/adl.ts
@@ -1,61 +1,43 @@
 /**
- * ADL Service — PERC-8293 (T11): Auto-Deleverage crank loop
+ * ADL Service — OBSERVABILITY ONLY.
  *
- * Originally scaffolded as PERC-8276. Updated in T11 to add:
- *   - Insurance fund utilization BPS threshold check
- *   - Two-phase dispatch comments / scaffolding for T5 (PERC-8270)
+ * Alignment with dcccrypto/percolator-prog `main` (verified against src/percolator.rs):
+ *   `ExecuteAdl` (tag 50) is ADMIN-GATED on-chain. `handle_execute_adl`:
+ *     - percolator.rs:14537  require_admin(header.admin, accounts[0])     → signer MUST be the market admin (else EngineUnauthorized 0xf)
+ *     - percolator.rs:14590  if insurance_fund.balance != 0 → InsuranceFundNotDepleted (0x2f)
+ *     - percolator.rs:14604  if max_pnl_cap > 0 && pnl_pos_tot <= cap → InvalidArgument
  *
- * Feature-flagged via env var `ADL_ENABLED=true` so it can run alongside
- * the existing crank service without affecting production behaviour until
- * the on-chain instruction is live.
+ *   The keeper's crank key is an unprivileged operational hot wallet, NOT the market
+ *   admin — so the keeper MUST NOT send ExecuteAdl (every send would revert), and we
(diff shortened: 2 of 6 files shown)


dcccrypto:
Closing - clean rebase applied via new PR with ADL observe-only, monitor.ts insuranceDepleted guard, and obsoletes PR #111.

PR #138, opened 2026-06-09
fix(keeper): make HA leader lock split-brain-safe via atomic CAS

### LINKED ISSUE #137: HA leader lock can split-brain: blind SET ... XX renew lets a stalled leader steal the lock back
## Summary

The HA leader lock (`src/lib/leader.ts`) can let two nodes hold leadership at the same time. `_renew()` refreshes the lock with a blind `SET key value XX` (set-if-exists) that does **not** verify the stored value is still this node's identity, and `stop()` deletes the lock unconditionally. A leader that stalls past the TTL loses the lock to a standby, then silently steals it back on its next renew — and both nodes then believe they are leader indefinitely.

## Mechanism

1. Node A holds the lock (`value = A`, TTL 30s) and renews every ~10s.
2. A stalls (GC pause, event-loop block, or a brief Redis blip) long enough that the TTL lapses and Redis evicts the key.
3. Node B (standby) polls, finds the key free, and acquires it with `SET … NX` (`leader.ts` poll path). B is now the legitimate leader (`value = B`).
4. A resumes. Its `_role` is still `"leader"` (it never demoted). Its renew runs `SET key A XX`; the key **exists** (held by B), so `XX` succeeds and **overwrites the value back to A**, returning `"OK"`. A stays leader.
5. A and B now both report leader and ping-pong the value on each renew — permanent split-brain.

Separately, `stop()` does an unconditional `DEL` when `_role === "leader"`, so a stale leader's graceful shutdown deletes a lock another node currently holds.

## Why it matters

The keeper issues **permissionless** on-chain instructions (`KeeperCrank`, `LiquidateAtOracle`) that the program does not gate by keeper identity, so two leaders produce duplicate liquidations/cranks, wasted priority fees, and failed-tx noise. The Redis lock is the only single-writer guarantee.

## Affected code

- `src/lib/leader.ts` — `_renew()` blind `SET … XX`; `stop()` unconditional `DEL`.
- `src/lib/redis-client.ts` — the `RedisLike` contract exposes only `set`/`get`/`del`, with no atomic compare-and-set primitive.

Related single-writer gaps in the same area: an involuntary demote does not drain/cancel the in-flight `sharedTxQueue`; `keeperSend` has no leadership re-check at send time; and `_renew` has no per-call timeout, so a hung Redis can strand a stale leader past the TTL.

## Reproduction (PoC)

Dependency-free; drives the real `LeaderLock` against an in-memory Redis that honors the `nx`/`xx`/`ex` semantics of the `RedisLike` contract. Reproduces against current `main`.

```ts
import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";

  createLogger: () => ({ info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() }),
}));

import { LeaderLock } from "../../src/lib/leader.js";
import type { RedisLike } from "../../src/lib/redis-client.js";

class FakeRedis implements RedisLike {
  store = new Map<string, string>();
  async set(key: string, value: string, opts: { ex: number; nx?: true } | { ex: number; xx?: true }) {
    const exists = this.store.has(key);
    if ("nx" in opts && opts.nx && exists) return null;
    if ("xx" in opts && opts.xx && !exists) return null;
    this.store.set(key, value);
    return "OK" as const;
  }
  async get(key: string) { return this.store.has(key) ? this.store.get(key)! : null; }
  async del(...keys: string[]) { let n = 0; for (const k of keys) if (this.store.delete(k)) n++; return n; }
  /** Simulate the lock's TTL lapsing while a node stalls. */
  expire(key: string) { this.store.delete(key); }
}

const OPTS = { ttlMs: 30_000, renewMs: 10_000, pollMs: 5_000 };
const LOCK_KEY = "keeper:leader:test";

describe("leader-lock split-brain via blind SET ... XX renew", () => {
  let redis: FakeRedis;
  beforeEach(() => { vi.useFakeTimers(); redis = new FakeRedis(); });
  afterEach(() => { vi.useRealTimers(); });

  it("a stalled-then-resumed leader steals the lock back — BOTH nodes end up leader", async () => {
    const demoteA = vi.fn();
    const nodeA = new LeaderLock(redis, "node-A", OPTS);
    const nodeB = new LeaderLock(redis, "node-B", OPTS);

    nodeA.start({ network: "test", onPromote: () => {}, onDemote: demoteA });
    await vi.advanceTimersByTimeAsync(0);
    expect(nodeA.role()).toBe("leader");

    nodeB.start({ network: "test", onPromote: () => {}, onDemote: () => {} });
    await vi.advanceTimersByTimeAsync(0);
    expect(nodeB.role()).toBe("standby");

    // A stalls; its TTL lapses (Redis evicts the key) before its renew fires.
    redis.expire(LOCK_KEY);

    // B's poll legitimately acquires the lock.
    await vi.advanceTimersByTimeAsync(OPTS.pollMs);
    expect(nodeB.role()).toBe("leader");
    expect(redis.store.get(LOCK_KEY)).toBe("node-B");

    // A resumes; `SET key node-A XX` succeeds against B's key and overwrites it.
    await vi.advanceTimersByTimeAsync(OPTS.renewMs - OPTS.pollMs);

    // SPLIT-BRAIN: both nodes report leader; A was never demoted and stole the value.
    expect(nodeA.role()).toBe("leader");
    expect(nodeB.role()).toBe("leader");
    expect(demoteA).not.toHaveBeenCalled();
    expect(redis.store.get(LOCK_KEY)).toBe("node-A");
  });

  it("stop() on a stale leader deletes a lock another node owns", async () => {
    const nodeA = new LeaderLock(redis, "node-A", OPTS);
    nodeA.start({ network: "test", onPromote: () => {}, onDemote: () => {} });
    await vi.advanceTimersByTimeAsync(0);
    expect(nodeA.role()).toBe("leader");

    // A's TTL lapsed and node B legitimately took the lock.
    redis.store.set(LOCK_KEY, "node-B");

    // A is still a stale leader; its graceful stop unconditionally deletes the key.
    await nodeA.stop();
    expect(redis.store.get(LOCK_KEY) ?? null).toBeNull(); // B's lock was wiped
  });
});
```

## Proposed fix

Make renew and release **atomic compare-and-set** operations, executed server-side via `@upstash/redis`'s `eval` (Lua): renew refreshes the TTL only if the stored value is still this node's identity (`pexpire`, value untouched), and release deletes only if the value is still this node's identity. A CAS "not owner" result demotes the node immediately. Add a per-call renew timeout so a hung Redis demotes instead of stranding a stale leader, a host-local single-writer gate in `keeperSend` (refuse to send when not leader), and a queue drop on involuntary demote. No fencing token — the program is permissionless, so there is no on-chain validator for one; the CAS lock plus the host-local send gate is the achievable single-writer guarantee.

A PR with a regression test follows.

0x-SquidSol:
## Summary

Fixes #137. Makes the HA leader lock split-brain-safe by replacing the blind `SET … XX` renew and unconditional `DEL` with atomic compare-and-set operations, and adds a host-local single-writer gate so a demoted node cannot land on-chain transactions.

## Root cause

`_renew()` renewed the lock with a blind `SET key value XX` that succeeds against any existing holder, so a leader that stalled past the TTL stole the lock back on its next renew and both nodes stayed leader. `stop()` deleted the lock unconditionally. The keeper sends permissionless `KeeperCrank`/`LiquidateAtOracle`, so dual leadership means duplicate liquidations/cranks; the Redis lock is the only single-writer guarantee.

## Changes

- **`src/lib/redis-client.ts`** — add `eval` to the `RedisLike` contract (mirrors `@upstash/redis` `Redis.eval`; the real client already satisfies it).
- **`src/lib/leader.ts`** — renew/release are now atomic compare-and-set Lua scripts: renew via `pexpire` **only if** `value == identity` (the value is never rewritten), release via `del` **only if** `value == identity`. A CAS "not owner" result demotes immediately (definitive loss); transport errors/timeouts keep the existing 2-strike tolerance. Adds a per-call renew timeout (`renewTimeoutMs`, default `min(renewMs/2, 5000)`) so a hung Redis demotes the node before its TTL silently lapses.
- **`src/lib/keeper-send.ts`** — `setLeaderCheck()` plus a first-thing single-writer gate: `keeperSend` refuses to send when the node is not leader and returns `null` (the existing "skip, not a failure" contract every caller already handles). This is the actual barrier against a demoted node double-sending, since the on-chain programs are permissionless.
- **`src/lib/tx-queue.ts`** — `clearPending()` drops queued-but-not-started tasks across all lanes.
- **`src/index.ts`** — wires the live-role send gate (`() => leaderLock ? role() === "leader" : true`) and clears the queued backlog on involuntary demote.
- **tests** — the split-brain reproduction is flipped into a regression (the stalled node now demotes with `redis-lock-lost`, the lock is not stolen, and stale `stop()` no longer deletes another node's lock), plus tests for renew-timeout demotion, CAS-renew (never a blind `SET XX`), the send gate, and `clearPending()`.

## Verification

- `tsc --noEmit`: clean.
- Full test suite: **619 passed**, 0 failed.
- Leader chaos/failover suite (`STRESS=true`): **18 passed** (graceful + ungraceful failover, network-partition standby, concurrent-acquire).

## Notes

- No behavior change off the HA path: the send gate defaults to "leader" when no lock is configured, and acquire still uses `SET … NX` unchanged.
- No fencing token: the program is permissionless, so there is no on-chain validator for a token; the CAS lock plus the host-local send gate is the achievable single-writer guarantee. (On-chain fencing would require a program change and is out of scope.)

## Release Notes

* **New Features**
  * HA mode now enforces leadership-aware transaction sending—non-leader nodes skip transactions and automatically clear pending queues on demotion.
  * Leader election uses atomic operations to prevent split-brain conflicts and stale leaders overwriting newer locks.

* **Bug Fixes**
  * Improved timeout handling for leader operations with graceful demotion on failures.

* **Tests**
  * Added comprehensive test coverage for leadership-aware transaction behavior and split-brain prevention scenarios.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,6 +19,7 @@ import { LeaderLock, makeIdentity } from "./lib/leader.js";
 import { captureAndExit } from "./lib/exit-handlers.js";
 import { StartupTracker } from "./lib/startup-tracker.js";
 import { sharedTxQueue, DRAIN_TIMEOUT_MS } from "./lib/tx-queue.js";
+import { setLeaderCheck } from "./lib/keeper-send.js";
 import { initSharedShadowHarness, sharedShadowHarness } from "./lib/shadow-harness.js";
 import { sharedDecisionLog } from "./lib/decision-log.js";
 
@@ -82,6 +83,10 @@ const leaderLock: LeaderLock | null =
 if (haEnabled && redisClient === null) {
   logger.warn("HA_ENABLED=true but KEEPER_REDIS_URL is unset — running as standalone leader");
 }
+
+// Single-writer guard for keeperSend: only land on-chain txs while we hold
+// leadership. Reads the live LeaderLock role (standalone / no-HA → always leader).
+setLeaderCheck(() => (leaderLock ? leaderLock.role() === "leader" : true));
 
 // A5: gate /health on real readiness — Railway otherwise marks the container
 // healthy the moment the HTTP server binds, well before start() finishes
@@ -681,6 +686,10 @@ async function start() {
       },
       onDemote: (reason) => {
         logger.warn("HA: demoted from leader — stopping services", { network, reason });
+        // role() is already "standby" here (LeaderLock flips it before onDemote),
+        // so the keeperSend single-writer guard is closed. Also drop any queued
+        // sends so this node runs no backlog after losing leadership.
+        sharedTxQueue.clearPending();
         stopAllServices();
       },
     });
--- a/src/lib/keeper-send.ts
+++ b/src/lib/keeper-send.ts
@@ -11,6 +11,16 @@ import { sharedDecisionLog } from "./decision-log.js";
 
 const logger = createLogger("keeper:send");
 
+// Single-writer guard. In HA mode the keeper must only land on-chain txs while it
+// holds leadership; the on-chain programs are permissionless, so this host-local
+// gate is the actual barrier against a demoted node double-sending. Wired from
+// index.ts to read the live LeaderLock role. Defaults to always-leader so the
+// standalone (no-HA) path is unchanged.
+let _isLeader: () => boolean = () => true;
+export function setLeaderCheck(fn: () => boolean): void {
+  _isLeader = fn;
+}
+
 export const BASE_FEE_LAMPORTS = 5_000;
 
 const TIER_MAP: Record<TxType, PriorityFeeTier> = {
@@ -111,6 +121,15 @@ export async function keeperSend(
   maxRetries = 3,
   keeperOpts?: KeeperSendOptions,
 ): Promise<KeeperSendResult | null> {
+  // Single-writer guard (checked first, before any RPC). A node that lost
+  // leadership — including a task dequeued from sharedTxQueue after demotion —
+  // must not land a tx. Returns null, reusing the budget-exhausted "skip, not a
+  // failure" contract every caller already handles.
+  if (!_isLeader()) {
+    logger.warn("Single-writer guard: refusing send — node is not leader", { txType });
+    return null;
+  }
+
   const { estimatedCost, simulatedCu } = await estimateCost(connection, instructions, signers, txType);
 
   if (!budget.canSpend(estimatedCost, txType)) {
--- a/src/lib/leader.ts
+++ b/src/lib/leader.ts
@@ -6,10 +6,26 @@ const logger = createLogger("keeper:leader");
 
 export type LeaderRole = "leader" | "standby" | "starting";
 
+/**
+ * Atomic compare-and-set scripts. The lock VALUE is this node's identity, so
+ * these only mutate the key when the caller still owns it — closing the
+ * split-brain window where a stalled-then-resumed leader's blind `SET XX` renew
+ * (or unconditional `DEL`) would clobber a lock a standby legitimately took.
+ *
+ * Renew uses `pexpire` (refresh TTL, value untouched) so even renew can never
+ * rewrite a foreign value. Exported so test fakes execute the exact same logic.
+ */
+export const RENEW_SCRIPT =
+  "if redis.call('get', KEYS[1]) == ARGV[1] then redis.call('pexpire', KEYS[1], ARGV[2]); return 1 else return 0 end";
+export const RELEASE_SCRIPT =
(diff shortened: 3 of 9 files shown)


dcccrypto:
Closing in favor of a clean rebase. CAS eval already landed in #126; remaining parts (setLeaderCheck, clearPending) applied via new PR.

PR #178, opened 2026-06-13
fix(liquidation): add a minimum-notional floor so the keeper skips dust

0x-SquidSol:
## Summary

The liquidation scanner has no minimum-notional floor: `scanMarket()` flags **any** account with `marginRatioBps < maintenanceMarginBps` and `notional > 0n` as a candidate, and the pre-submit re-check applies the same test. The keeper liquidates as a permissionless cranker and earns **no** liquidation fee (the on-chain penalty is charged to the liquidated trader and goes to the protocol, not the keeper), so every liquidation is a pure cost — a crank+`LiquidateAtOracle` tx paying a Jito tip (default 200,000 lamports) plus fees. So the keeper spends real SOL liquidating dust worth less than the fee, and — because the program allows dust deposits (no per-deposit minimum) — an attacker can open many tiny undercollateralized positions across many owner accounts (bypassing the per-owner cap) to drain the wallet up to the budget caps, at which point the budget breaker halts the keeper entirely.

## Fix

Add a configurable minimum-notional floor (`KEEPER_MIN_LIQ_NOTIONAL`, default **$1**) and skip candidates below it in **both** `scanMarket()` and the pre-submit re-check in `liquidate()` — so a position that slips through the scan, or the event-driven path that calls `liquidate()` directly, is still floored before paying a fee.

### Why a single flat floor is decimal-correct

`notional = abs(positionSize) * priceE6 / 1e6`. `POS_SCALE` is a fixed `1_000_000` (program spec) and `priceE6` is a USD price scaled by `1e6`, so `notional` is denominated in a **fixed 6-decimal USD quote unit** (`1_000_000n == $1`) for every market — the collateral mint's token decimals (SOL 9, wBTC 8, …) do **not** enter it. So one USD-denominated floor is correct across all markets; no per-mint configuration is needed.

The default ($1) is ~10–20× the per-liquidation fee yet far below any economically meaningful position. The dangerous failure mode for a floor is "too high → skip a real liquidation", so the default is deliberately conservative; operators tune via `KEEPER_MIN_LIQ_NOTIONAL` (raw 6-decimal-USD units) and can set it to `0` to disable.

### Residual exposure

Sub-floor positions are left unliquidated, but the residual bad debt per position is immaterial (< the floor) and is backstopped by the insurance fund / ADL. Parking aggregate exposure under the floor is uneconomic for an attacker (each dust position costs on-chain rent + fees + forfeited collateral to open, while the keeper pays nothing to skip). A new metric, `keeper_liquidation_skipped_dust_total` (labelled by stage `scan`|`presubmit`), surfaces dust-spam activity so operators can watch aggregate skipped exposure rather than relying on an automated cap (which an attacker could weaponize into a halt).

## Tests

Adds a regression test: a dust ($0.01) position is skipped while a material ($10,000) one is kept; a position exactly at the floor is kept and one just below is skipped; the pre-submit path aborts (returns `null`, no send) on a sub-floor fresh notional. Plus parser unit tests for `KEEPER_MIN_LIQ_NOTIONAL` (default / valid / invalid / negative / explicit-zero). Existing `scanMarket`/`liquidate` tests use $10,000 positions and are unaffected.

Full suite: 694 passing, 25 skipped; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **New Features**
  * Liquidations below a configurable minimum notional threshold (default $1 USD) are now automatically filtered out during market scanning and pre-submit validation stages.

* **Chores**
  * Added monitoring metrics to track skipped liquidations due to notional floor enforcement.

### DIFF
--- a/src/lib/metrics.ts
+++ b/src/lib/metrics.ts
@@ -26,6 +26,13 @@ export const solSpentLamportsTotal = new Counter({
   registers: [registry],
 });
 
+export const liquidationSkippedDustTotal = new Counter({
+  name: "keeper_liquidation_skipped_dust_total",
+  help: "Liquidation candidates skipped for being below the min-notional floor, partitioned by stage (scan|presubmit)",
+  labelNames: ["stage"] as const,
+  registers: [registry],
+});
+
 export const jitoBundleFailCountTotal = new Counter({
   name: "keeper_jito_bundle_fail_count_total",
   help: "Total number of Jito bundle submission failures",
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -24,6 +24,7 @@ import {
   solSpentLamportsTotal,
   cycleDurationSeconds,
   txLandTimeSeconds,
+  liquidationSkippedDustTotal,
 } from "../lib/metrics.js";
 import type { AccountLoader } from "../lib/account-loader.js";
 import { keeperSend, sharedBudget } from "../lib/keeper-send.js";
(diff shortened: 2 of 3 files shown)


dcccrypto:
This fix has already been applied to main (the min-notional floor, parseMinLiqNotional, KEEPER_MIN_LIQ_NOTIONAL env var, metrics counter, and tests are all present in the current HEAD). Closing as already incorporated.

PR #180, opened 2026-06-13
fix(oracle): resolve Chainlink oracle accounts by owner instead of always deriving a Pyth PDA

0x-SquidSol:
## Summary

The keeper had no oracle-type discriminator for non-HYPERP markets: crank, liquidation, and ADL all derived a Pyth Push PDA for the oracle account. For a **Chainlink**-oracle market this is the wrong account, so every crank, liquidation, and ADL on that market reverts on-chain and the market is never serviced — it stales, accrues consecutive failures, and is eventually paused/evicted.

## Root cause

The program (`percolator.rs` `read_engine_price_e6`) detects the oracle type by the supplied oracle account's **owner** — `PYTH_RECEIVER_PROGRAM_ID` → Pyth, `CHAINLINK_OCR2_PROGRAM_ID` → Chainlink, else `IllegalOwner`. For Chainlink, `config.index_feed_id` **is** the aggregator account pubkey (the program checks `price_ai.key == index_feed_id`); for Pyth, `index_feed_id` is the feed id and the oracle account is the derived Pyth Push PDA. The two are indistinguishable from config alone (both `index_feed_id != 0`, post-Phase-G `oracle_authority == 0`).

The keeper derived a Pyth PDA for every non-HYPERP market (`crank.ts`, `liquidation.ts`, `adl.ts`), so a Chainlink market got an account owned by neither program with key ≠ `index_feed_id` → `IllegalOwner`/`InvalidOracleKey` → permanent revert. Chainlink is a supported, deployable oracle (`CHAINLINK_OCR2_PROGRAM_ID` is a real mainnet program; `InitMarket` reads through the same dispatch).

## Fix

Add a shared resolver, `src/lib/oracle-account.ts`, that distinguishes Pyth from Chainlink by the **on-chain owner** of the account at `index_feed_id`, cached per feed (the owner is fixed at `InitMarket`):

- owner == `CHAINLINK_OCR2` → oracle account = `index_feed_id` (the aggregator)
- otherwise / null / RPC error → derive the Pyth Push PDA (pre-fix behavior)

This is a **strict correctness superset** of the old code: a Pyth market is byte-for-byte unchanged; a flaky owner lookup never makes a working market worse; a confirmed Chainlink owner routes correctly. A confirmed verdict is cached (no per-send RPC); an inconclusive/errored lookup is **not** cached, so a Chainlink market self-heals on the next send once the lookup succeeds.

The three services keep their existing **HYPERP / admin pre-checks (→ slab) unchanged** and call the resolver only in the external-feed branch — so HYPERP and Pyth behavior is provably unchanged and the Pyth-vs-Chainlink logic lives in exactly one place (eliminating the three-way drift that caused the bug). `CHAINLINK_OCR2_PROGRAM_ID` is pinned locally (the SDK does not export it) with a unit test asserting the literal.

## Tests

- `tests/lib/oracle-account.test.ts` — resolver unit tests: Chainlink owner → `index_feed_id`; null/unrelated owner → Pyth PDA; positive verdict cached (one `getAccountInfo` across repeated resolves); RPC error → Pyth fallback **not** cached (self-heals on retry); the pinned program-id literal.
- `tests/services/chainlink-oracle-account.test.ts` — end-to-end through `liquidate()`: a Chainlink market passes `index_feed_id` (the aggregator) into the `LiquidateAtOracle` instruction; a Pyth market (feed id is not an account) derives the Pyth Push PDA.
- Updated two existing test mocks (`adl`, `crank-error-code`) to provide the feed's `toBase58` + a `getAccountInfo` the resolver now consults.

Full suite: 696 passing, 25 skipped; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **New Features**
  * Added automatic oracle provider detection for market operations, enabling seamless handling across different oracle types
* **Tests**
  * Introduced comprehensive test coverage for oracle account resolution and market compatibility scenarios

### DIFF
--- a/src/lib/oracle-account.ts
+++ b/src/lib/oracle-account.ts
@@ -1,0 +1,94 @@
+import { PublicKey } from "@solana/web3.js";
+import type { Connection } from "@solana/web3.js";
+import { derivePythPushOraclePDA } from "@percolatorct/sdk";
+import { createLogger } from "@percolatorct/shared";
+
+const logger = createLogger("keeper:oracle-account");
+
+/**
+ * Chainlink OCR2 Store program id.
+ *
+ * The on-chain program (percolator-prog `read_engine_price_e6`) detects the
+ * oracle type by the supplied oracle account's OWNER — PYTH_RECEIVER → Pyth,
+ * CHAINLINK_OCR2 → Chainlink — and for a Chainlink market `index_feed_id` IS the
+ * aggregator account pubkey (the program checks `price_ai.key == index_feed_id`).
+ * The SDK does not export this constant, so it is pinned here against the
+ * program's `CHAINLINK_OCR2_PROGRAM_ID` (percolator.rs). A unit test asserts the
+ * literal so a drift is caught loudly.
+ */
+export const CHAINLINK_OCR2_PROGRAM_ID = new PublicKey(
+  "HEvSKofvBgfaexv23kMabbYqxasxU3mQ4ibBMEmJWHny",
+);
+
+/** Minimal feed shape — compatible with a real PublicKey and parsed-config mocks. */
+interface FeedLike {
+  toBytes(): Uint8Array;
+  toBase58(): string;
+}
+
+type OracleKind = "chainlink" | "pyth";
+
+/** feed pubkey (base58) → resolved oracle kind. The owner of a feed account is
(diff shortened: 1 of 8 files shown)


dcccrypto:
Closing in favor of a clean rebase against current main. The fix has been rebased and a new PR will be created.

PR #194, opened 2026-06-14
fix(crank): M8 — add per-market in-flight guard so registerMarket cannot race the timer cycle

6figpsolseeker:
## Summary

Fixes **M8** from the internal pre-mainnet audit. Adds a per-market in-flight guard to `crankMarket` so concurrent callers (e.g., `registerMarket` via the `/register` HTTP endpoint and the timer-driven `crankAll` fan-out) can't produce duplicate KeeperCrank txs for the same slab.

## Root cause

`CrankService.registerMarket` at `src/services/crank.ts:1222` calls `await this.crankMarket(slabAddress)` directly — bypassing the process-wide `_cycling` guard that gates only the setInterval-driven `crankAll` loop. Any path that reaches `crankMarket` outside that loop (`/register` HTTP, LaserStream debounce handlers, etc.) slips past it.

A `/register` HTTP arriving mid-cycle can produce **two concurrent `crankMarket` calls for the same slab**, each building a fresh blockhash and signing independently. Solana's signature-based dedup does NOT apply (different signatures). Result: duplicate `KeeperCrank` txs, wasted priority fees, possibly doubled funding accrual within the same window if the on-chain handler is not slot-gated.

## Fix

Per-market in-flight set:

```ts
// field
private _inflightMarkets = new Set<string>();

async crankMarket(slabAddress: string): Promise<boolean> {
  if (this._inflightMarkets.has(slabAddress)) {
    logger.debug("crankMarket: in-flight for slab, skipping concurrent call", { slabAddress });
    return false;
  }
  // ...existing market/state lookup...
  this._inflightMarkets.add(slabAddress);
  try {
    // ...existing body...
  } catch (err) {
    // ...existing error handling...
    return false;
  } finally {
    this._inflightMarkets.delete(slabAddress); // released even on `return false` inside try/catch
  }
}
```

No changes needed to `registerMarket` itself — its call to `crankMarket` now respects the same guard automatically.

This also closes the deferred **Option F defense-in-depth** that Agent C recommended during the **H4** (#132) hostile reverify — a per-market guard alongside the watchdog fix.

## Files touched

- `src/services/crank.ts` — add `_inflightMarkets` field, guard check at entry, `finally { delete }` clause.
- `tests/services/crank.test.ts` — 2 new unit tests.
- `tests/poc/m8.poc.test.ts` — 5 PoC tests.

## Test plan

- [x] **`crankMarket bails when slab is already in _inflightMarkets`** — pre-populate the set, call crankMarket, verify it returns `false` and `keeperSend` is not called.
- [x] **`_inflightMarkets is empty before AND after a crankMarket call`** — verify the `finally` clause always releases.
- [x] **PoC**: OLD pattern (no guard) — two concurrent calls each send a tx (DUPLICATE).
- [x] **PoC**: NEW pattern — second concurrent call returns false, only ONE tx sent.
- [x] **PoC**: Per-slab isolation — different slabs don't block each other.
- [x] **PoC**: Guard releases on body throw.
- [x] **PoC**: Sequential calls work normally.
- [x] crank suite: 28 passed (+2 new M8 tests).
- [x] M8 PoC: 5 passed.
- [x] Full vitest suite: **619 passed / 26 skipped** — no regressions.
- [x] `pnpm build` clean.

## Severity

**MEDIUM** — real concurrency hole. Orthogonal to H4 (#132): H4 was the watchdog-clear race; M8 is the bypass-entry race. Both live in audit cluster `CL-DOUBLE-EXEC`. With both fixes in place, `crankMarket` can never run concurrently for the same slab from any code path.

## Risk

- **Backward-compatible.** Sequential calls (normal operation) are unaffected — the guard is per-call, not permanent.
- **Defense-in-depth.** Even after H4 lands, the per-market guard catches any future code path that calls `crankMarket` outside the timer loop.

## Related

- **H4** (#132) — watchdog fix that prevented `_cycling` flag flipping while `crankAll` was in-flight.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed concurrent operation handling to prevent duplicate transactions from being submitted when multiple calls target the same market simultaneously.

* **Tests**
  * Added test coverage verifying the concurrent operation guard works correctly across various scenarios, including error conditions and sequential operations.

### DIFF
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -183,6 +183,16 @@ export class CrankService {
   private _isRunning = false;
   private _cycling = false;
   private _cycleStartedAt = 0;
+  // M8: per-market in-flight guard. `_cycling` is a process-wide flag for
+  // the timer-driven crankAll cycle, but other entry points (registerMarket
+  // from the /register HTTP endpoint, and potentially LaserStream debounce
+  // handlers) can call crankMarket directly — bypassing `_cycling`. Without
+  // this guard, a /register HTTP arriving mid-cycle could fire crankMarket
+  // concurrently with crankAll's fan-out on the same slab, producing
+  // duplicate KeeperCrank txs for the same market. Set/delete at crankMarket
+  // entry/exit so every code path that reaches crankMarket honors the same
+  // in-flight invariant.
+  private _inflightMarkets = new Set<string>();
   private _stalePauseCheck?: (slabAddress: string) => boolean;
   // P1 FIX: Cache keypair at construction — was reading from disk on every crank cycle (every 30s)
   private readonly _keypair = loadKeypair(process.env.CRANK_KEYPAIR!);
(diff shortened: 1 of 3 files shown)


dcccrypto:
Superseded by #207 (rebased onto main, conflict resolved).

PR #203, opened 2026-06-14
Close two budget circuit-breaker bypasses: NaN cost slips canSpend, and concurrent sends overshoot caps (TOCTOU)

### LINKED ISSUE #198: Budget circuit-breaker can be silently bypassed or overshot — NaN cost slips canSpend, and concurrent sends overshoot caps (TOCTOU)
## Summary

The keeper budget is the circuit breaker that bounds send-path SOL spend (per cycle / hour / day) and halts until an operator resumes. Two gaps let it be silently bypassed or overshot.

### 1. A non-finite cost slips through `canSpend()` and is never recorded → breaker silently off

`estimateCost()` builds `estimatedCost` from a priority fee + CU + `parseInt(process.env.JITO_TIP_LAMPORTS ?? "200000", 10)`. A malformed env (`JITO_TIP_LAMPORTS="abc"`) makes the tip — and therefore `estimatedCost` — `NaN`.

- `canSpend(NaN)` passes every cap check, because `NaN > cap` is `false` for all of them, and returns `true`.
- `recordTx(NaN, …)` early-returns on `!Number.isFinite(lamports)` and records nothing.

So with one malformed fee env, **every send is admitted and nothing is ever recorded** — the cycle/hour/day caps never trip and the breaker is silently disabled for the life of the process. A misconfiguration that should fail loudly instead removes the wallet's last line of defense.

### 2. `canSpend()` TOCTOU → concurrent sends overshoot the caps

`canSpend()` is a pure read: it checks the caps but books nothing. The spend is only recorded by `recordTx()`, which runs in the send's `finally` — after the network `await`. The keeper sends concurrently (the tx-queue lanes run up to 10 liquidation + 5 oracle + 5 crank in flight at once). Every in-flight send calls `canSpend()` against the *same* pre-send snapshot, so they all pass and collectively spend past the cap before any of them records. The class comment ("no two calls can interleave between their reads and writes") is true per-method, but the gap here is *across* methods, straddling the await.

The overshoot is bounded by `(in-flight concurrency) × (per-tx cost)`, so it is not an unbounded drain — but the caps are the manual-resume fund-safety backstop and should be hard, not soft.

## Reproduction (PoC)

Both fail on current `main`:

```ts
import { KeeperBudget } from "../../src/lib/budget.js";

const CFG = { maxSolPerCycle: 100, maxSolPerHour: 1e9, maxSolPerDay: 1e9, maxTxPerCycle: 1000 };

// (1) NaN cost bypass
const b1 = new KeeperBudget(CFG, { now: () => 1 });
expect(b1.canSpend(NaN, "crank")).toBe(false); // FAILS on main: returns true
expect(b1.isHalted()).toBe(true);              // FAILS on main: not halted

// (2) canSpend TOCTOU — two 60-lamport sends in flight, cap 100
const b2 = new KeeperBudget(CFG, { now: () => 1 });
expect(b2.canSpend(60, "crank")).toBe(true);
expect(b2.canSpend(60, "crank")).toBe(false);  // FAILS on main: returns true (nothing reserved yet)
```

## Proposed fix

- **Non-finite cost:** treat a non-finite `lamports` in `canSpend()` as a breach — refuse the send and halt (fail-safe), so a bad fee env surfaces immediately and an operator must fix it and `resume()`. Validate `JITO_TIP_LAMPORTS` (and any numeric fee env) at boot so the `NaN` never reaches the hot path in the first place.
- **TOCTOU:** have `canSpend()` reserve the proposed cost into a pending tally that the cap checks include, so concurrent in-flight sends see each other; `recordTx()` releases the reservation (clamped at zero) and books the settled amount. The existing recorded-spend accounting (`cycleSpend` = sum of non-drop `recordTx`) is left unchanged, so the standalone `recordTx` contract and its property tests still hold.

A PR with the fix and tests (the PoCs above plus reservation-release and property coverage) follows.

0x-SquidSol:
## Problem

The `KeeperBudget` circuit breaker caps send-path SOL spend (per cycle / hour / day + tx count) and halts until an operator resumes. Two gaps let it be silently bypassed or overshot.

**1. A non-finite cost slips `canSpend()` and is never recorded → breaker silently off.**
`estimateCost()` builds `estimatedCost` from a priority fee + CU + `parseInt(JITO_TIP_LAMPORTS ?? "200000")`. A malformed env (`"abc"`) makes the tip — and `estimatedCost` — `NaN`. `canSpend(NaN)` passes every cap check (`NaN > cap` is `false`) and returns `true`; `recordTx(NaN)` early-returns on `!Number.isFinite`. So one malformed env admits every send and records nothing — the caps never trip for the life of the process.

**2. `canSpend()` TOCTOU → concurrent sends overshoot the caps.**
`canSpend()` only reads the counters; spend is booked by `recordTx()` in the send's `finally`, after the network `await`. The tx-queue runs up to 10 liquidation + 5 oracle + 5 crank sends in flight; each calls `canSpend()` against the same pre-send snapshot, so they all pass and collectively spend past the cap before any records.

## Fix

**Non-finite cost (G):** `canSpend()` treats a non-finite/negative `lamports` as a breach — refuse and halt (fail-safe `non-finite-cost` kind), so a bad fee env surfaces immediately and requires an explicit `resume()`. `JITO_TIP_LAMPORTS` is also validated at boot (`validateKeeperEnvGuards`) so the `NaN` never reaches the hot path.

**TOCTOU (B):** `canSpend()` now reserves the proposed cost (and one tx slot) into a pending tally that the cap checks include, so concurrent in-flight sends see each other; `recordTx()` releases the reservation (clamped `>= 0`) and books the settled amount.
- A **settled-spend** breach (already-booked spend would exceed a cap) still **halts** — unchanged semantics, and all existing cap tests stay green.
- A breach driven **only by in-flight reservations** **refuses without halting** — concurrency back-pressure must not latch the breaker and stall the keeper during exactly the bursts (e.g. liquidation cascades) it exists to handle. Capacity returns automatically as in-flight sends settle.

**Why the property tests still hold:** reservations live in fields (`_reservedLamports` / `_reservedTxCount`) disjoint from `_cycleSpend`/`_hourSpendSum`/etc. `recordTx` only ever decrements them (clamped at 0), so a `recordTx`-only sequence (as the property tests use) leaves `getStats().cycleSpend` exactly equal to the sum of non-drop `recordTx` lamports.

**Leak-proofing (essential):** `keeperSend` is restructured so every reserving `canSpend` is paired with exactly one `recordTx` via an outer `try/finally` + an idempotent release closure. Without this, an unexpected throw between the gate and the send would leak a reservation, silently shrinking the effective cap until the budget wedged — worse than the TOCTOU it fixes. `beginCycle` deliberately does **not** reset reservations (in-flight sends straddle cycle boundaries and release on their own `recordTx`).

## Tests

- `tests/lib/budget-gate-accuracy.poc.test.ts` (new): the two PoCs (NaN refused+halts; second concurrent `canSpend` refused), both fail on `main`.
- `tests/lib/budget.test.ts`: non-finite halt (+ `onHalt` kind), settled-breach-still-halts vs reservation-refuses, reservation release on success/fail/drop, reserved-tx-count back-pressure, `beginCycle` does not orphan reservations.
- `tests/lib/keeper-send.test.ts`: reservation-release (no-leak) guards across success / throw / DRY_RUN / canSpend-refused / 100 interleaved sends.
- `tests/env-guards.test.ts`: `JITO_TIP_LAMPORTS` boot validation (non-numeric / negative / trailing-garbage rejected; valid & unset accepted).
- All 5 budget property-test invariants unchanged; full suite green (711 passed) and `tsc --noEmit` clean.

Fixes #198

### DIFF
--- a/src/env-guards.ts
+++ b/src/env-guards.ts
@@ -89,6 +89,27 @@ export function validateKeeperEnvGuards(env: NodeJS.ProcessEnv = process.env): v
     }
   }
 
+  // Validate JITO_TIP_LAMPORTS at boot. It feeds estimatedCost in keeper-send
+  // (and crank/liquidation/adl) via parseInt(... ?? "200000", 10), which the
+  // KeeperBudget circuit breaker gates on. A malformed value (e.g. "abc")
+  // parses to NaN; before the canSpend NaN guard that silently disabled the
+  // breaker, and even with it the keeper would halt on first send. Reject a
+  // set-but-malformed value at boot (a clean supervisor restart) rather than
+  // discover it mid-incident. Unset/empty is fine — the read sites default to
+  // 200000. Validate with Number() (not parseInt) so "200000abc" is rejected
+  // rather than silently truncated.
+  const jitoTipRaw = env.JITO_TIP_LAMPORTS?.trim();
+  if (jitoTipRaw !== undefined && jitoTipRaw !== "") {
+    const tip = Number(jitoTipRaw);
+    if (!Number.isFinite(tip) || !Number.isInteger(tip) || tip < 0) {
+      throw new Error(
+        `JITO_TIP_LAMPORTS='${jitoTipRaw.slice(0, 20)}' is invalid — expected a ` +
+          `non-negative integer (lamports). A malformed value makes the tx-cost ` +
+          `estimate NaN and would trip the keeper spend circuit breaker.`,
+      );
+    }
+  }
+
   const serviceRoleKey = env.SUPABASE_SERVICE_ROLE_KEY?.trim();
 
   // K-2 (HIGH): hard-reject SUPABASE_SERVICE_ROLE_KEY being present at all.
--- a/src/lib/budget.ts
+++ b/src/lib/budget.ts
@@ -54,6 +54,10 @@ export interface BudgetStats {
   hourSpend: number;
   daySpend: number;
   cycleTxCount: number;
+  /** Lamports reserved by in-flight canSpend()s not yet settled by recordTx(). */
+  reservedLamports: number;
+  /** Count of in-flight canSpend()s not yet settled by recordTx(). */
+  reservedTxCount: number;
   txSuccessRate: number | null;
   txWindowSize: number;
   halted: boolean;
@@ -68,6 +72,7 @@ export type HaltKind =
   | "day-spend-cap"
   | "cycle-tx-count-cap"
   | "tx-success-rate"
+  | "non-finite-cost"
   | "operator";
 
 export interface KeeperBudgetDeps {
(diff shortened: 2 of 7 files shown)


dcccrypto:
Superseded by #208 (rebased onto main).

Issue #150, opened 2026-06-09
H4: Crank watchdog flips _cycling while crankAll() is in-flight (intra-process double-execution)


6figpsolseeker:
## Summary

The setInterval watchdog in `CrankService.start()` set `_cycling = false` whenever elapsed cycle time exceeded MAX_CYCLE_MS. But the original `crankAll()` invocation was still awaiting Promise.all results — the watchdog never cancels anything, it just clears the in-flight guard. The next interval tick sees `_cycling === false` and launches a SECOND `crankAll()` concurrently.

## Severity

**HIGH** per Agent A, **MEDIUM** per Agents B+C. Real concurrency bug. Consequences: duplicate KeeperCrank txs per market, doubled funding accrual, doubled Jito tips, RPC status-poll storms. Cascading race: the original cycle's `finally { _cycling=false }` later clobbers the second cycle's `_cycling=true`, allowing a third to start.

## Location

`src/services/crank.ts:1264-1314` (pre-fix watchdog + finally block)

## Bug

```ts
if (this._cycling) {
  if (elapsed > MAX_CYCLE_MS) {
    sendCriticalAlert(...).catch(() => {});
    this._cycling = false; // ← THE BUG: in-flight crankAll() still running
  }
  return;
}
this._cycling = true;
// ...crankAll()...
```

`recentSignatures` is post-hoc tracking (doesn't gate sends). `isDue()` reads `lastCrankTime` which is only updated after a successful send — so during a hung first cycle, `isDue` returns true for every market on the second cycle.

## Threat model

Helius/Triton RPC latency spike → crankAll() takes >MAX_CYCLE_MS to complete its Sender retries. Watchdog flips `_cycling=false`. Next tick launches a parallel crankAll(). Both fan out to per-market `crankMarket()` calls with FRESH blockhashes (so Solana signature-dedup does NOT apply). Both submit. On-chain instructions process whichever lands first, but funding/queue side effects of the second tx are non-trivial.

## Fix

See #132. The watchdog now arms a 30s grace timer and `process.exit(1)`s if the cycle stays hung past grace. It NEVER flips `_cycling`, so no second concurrent crankAll() can launch. Natural recovery (cycle's `finally` block) disarms the watchdog cleanly. Process exit lets the supervisor (Railway/PM2/k8s) restart — the only honest "cancel" available without plumbing AbortController through the entire Sender stack.

[...]

PR #205, opened 2026-06-14
fix(crank): M9 — advance lastCrankTime on failure so the inactive back-off interval is honored

6figpsolseeker:
## Summary

Fixes **M9** from the internal pre-mainnet audit. Advances `state.lastCrankTime` on failure so the existing inactive back-off interval (120s) is actually honored after a market is deactivated.

## Root cause

`CrankService.isDue(state)` at `src/services/crank.ts:678`:

```ts
private isDue(state: MarketCrankState): boolean {
  const interval = state.isActive ? this.intervalMs : this.inactiveIntervalMs;
  return Date.now() - state.lastCrankTime >= interval;
}
```

After 10 consecutive failures the catch block at line ~952 flips `state.isActive = false`, intending to slow polls from `intervalMs` (30s active) to `inactiveIntervalMs` (120s inactive). But the catch block only updates `failureCount` and `consecutiveFailures` — never `lastCrankTime`:

```ts
} catch (err) {
  state.failureCount++;
  state.consecutiveFailures++;
  // ← lastCrankTime NOT updated
  // ...
  if (state.consecutiveFailures >= 10) {
    state.isActive = false; // back-off intent
  }
  // ...
}
```

So when the inactive phase kicks in:
- If the market never succeeded → `lastCrankTime = 0` → `isDue` always true.
- If the market succeeded long ago → `lastCrankTime` is stale → `isDue` always true.

Either way, the dead market keeps cranking every 30s (active interval) even after `isActive` flips to false. The intended 4× reduction in poll frequency never materializes. RPC quota + priority fees burn on broken markets indefinitely.

## Fix

Single line at the top of the catch block:

```ts
} catch (err) {
  state.failureCount++;
  state.consecutiveFailures++;
  state.lastCrankTime = Date.now(); // ← M9: advance the time anchor on failure too
  // ...
}
```

The success path already does this (line 905). After the fix:
- **Active failing markets** still retry every `intervalMs` (30s) — active interval is small, so the back-off doesn't kick in yet.
- **After 10 failures**, `isActive` flips. `lastCrankTime` is now the time of the last failure. Next `isDue` check uses `inactiveIntervalMs` (120s) against a fresh time anchor → correctly skips until 120s elapse.

**4× reduction in poll frequency for dead markets (30s → 120s).**

## Files touched

- `src/services/crank.ts` — single-line fix + multi-line rationale comment.
- `tests/services/crank.test.ts` — 1 new unit test.
- `tests/poc/m9.poc.test.ts` — 4 PoC tests demonstrating OLD vs NEW behavior.

## Test plan

- [x] **Unit**: `lastCrankTime` advances on every failed `crankMarket` attempt (verified with `vi.setSystemTime`).
- [x] **PoC**: OLD pattern — market never succeeds → no back-off, cranks every 30s.
- [x] **PoC**: NEW pattern — `lastCrankTime` advances → 120s back-off honored after 10 failures.
- [x] **PoC**: NEW pattern — active failing market unaffected (still 30s).
- [x] **PoC**: NEW pattern — recovery still resets cleanly to active cadence.
- [x] crank suite: **26 passed** (was 25 — +1 new M9 test).
- [x] M9 PoC: 4 passed.
- [x] Full vitest suite: **617 passed / 26 skipped** — no regressions.
- [x] `pnpm build` clean.

## Severity

**LOW** — operational hygiene / resource hygiene. No security or fund-loss path. The waste is bounded by `(failing markets) × (cycle frequency)` and is purely an RPC-quota / priority-fee concern.

## Related

- **PERC-381** / line 938 — the catch block already does exponential cooldown for the specific 0x4 error (`NotInitialized`) by setting `permanentlySkipped` + `permanentlySkippedAt`. M9 generalizes the back-off intent to ALL consecutive failures, not just 0x4.
- **H4** (#150) / **M8** (#194) — both addressed concurrency races in the crank loop. M9 addresses a separate resource-hygiene gap in the same file.

### DIFF
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -892,6 +892,19 @@ export class CrankService {
     } catch (err) {
       state.failureCount++;
       state.consecutiveFailures++;
+      // M9: advance lastCrankTime on FAILURE too, not just success. Pre-fix,
+      // `isDue` checks `Date.now() - state.lastCrankTime >= interval`. With
+      // lastCrankTime only updated on success, a market that's been failing
+      // every cycle has a stale lastCrankTime (or 0 if it never succeeded),
+      // so isDue is always true. After 10 consecutive failures isActive
+      // flips to false and the keeper SHOULD back off from intervalMs (30s)
+      // to inactiveIntervalMs (120s) — but since lastCrankTime is stale, the
+      // back-off never actually fires and the dead market keeps cranking
+      // every 30s, wasting RPC quota + priority fees.
+      // Advancing lastCrankTime here makes the inactive interval honored:
+      // active-failing markets still retry every intervalMs (30s);
+      // deactivated markets retry every inactiveIntervalMs (120s) as designed.
+      state.lastCrankTime = Date.now();
       txSentTotal.inc({ result: "fail", type: "crank" });
 
       const errMsg = err instanceof Error ? err.message : String(err);
--- a/tests/poc/m9.poc.test.ts
+++ b/tests/poc/m9.poc.test.ts
@@ -1,0 +1,151 @@
+/**
+ * M9 PoC — deactivated markets still poll at the active interval because
+ * lastCrankTime isn't advanced on failure.
+ *
+ * THE BUG (pre-fix):
+ *   `isDue(state)` at src/services/crank.ts:678 returns:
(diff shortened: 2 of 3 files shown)


dcccrypto:
Closed by #214 — rebased onto main and squash-merged. The M9 fix is now on main.

PR #212, opened 2026-06-14
fix(keeper-send): M12 — reconcile budget against realized on-chain cost

6figpsolseeker:
## Finding M12 — keeperSend records estimated, not realized, cost

**Severity:** Medium
**Cluster:** Observability / budget accuracy
**Files:** [src/lib/keeper-send.ts](src/lib/keeper-send.ts), [src/lib/budget.ts](src/lib/budget.ts)

### Problem

[src/lib/keeper-send.ts:204](src/lib/keeper-send.ts#L204) calls
`budget.recordTx(estimatedCost, txType, result)` in the `finally` block,
where `estimatedCost` was computed pre-send by
[`estimateCost()`](src/lib/keeper-send.ts#L71-L91) as:

```
estimatedCost = BASE_FEE + ceil(microLamports * simulatedCu / 1e6) + jitoTip
```

The `simulatedCu` value comes from the local CU estimator and can diverge
from the runtime cost — cold accounts, inner-CPI fanout and lookup-table
indirection all push real CU above the local sim. The budget therefore
**under-counts spend whenever the sim under-estimates CU**, and the gate
at [keeper-send.ts:116](src/lib/keeper-send.ts#L116) keeps approving sends
long after `maxSolPerCycle` has been blown.

### Fix

After a successful send, schedule a fire-and-forget reconciliation:

1. `scheduleRealizedCostReconciliation` (new) waits 5s for confirmed slot
   lag, calls `connection.getTransaction(signature, { commitment: 'confirmed' })`,
   reads `meta.fee`, and adds back the Jito tip when `USE_HELIUS_SENDER=true`
   so the realized total is comparable to `estimatedCost`.
2. New `KeeperBudget.adjustForRealizedCost(estimated, realized, txType)`:
   - shifts `cycleSpend / hourSpendSum / daySpendSum` by `(realized - estimated)`
   - **clamps each counter at 0** so a negative reconciliation never drives
     spend negative
   - tracks cumulative `realizedCostDriftLamports` and `realizedCostSamples`
     for telemetry / alerting
3. Errors are caught inside the IIFE — a flaky `getTransaction` never
   propagates back to the send path.
4. Sampling is env-controlled via `KEEPER_REALIZED_COST_SAMPLE_PCT`
   (default 100 — reconcile every tx). Deterministic on signature so the
   same tx is never double-sampled.

### Non-goals

- This is **observability + soft tightening**, not a hard halt. A blown
  estimate this tx tightens the budget for the next tx; it does not unwind
  a tx that already shipped.
- We do not change the `canSpend` gate semantics — `estimatedCost` is still
  the pre-send check.

### Tests

- [`tests/lib/budget.test.ts`](tests/lib/budget.test.ts) — 6 new unit
  tests covering: positive delta, negative delta, clamp-at-zero,
  NaN / non-finite / negative rejection, cumulative drift across 10 tx,
  and the gate-tightening behaviour.
- [`tests/lib/keeper-send.m12-poc.test.ts`](tests/lib/keeper-send.m12-poc.test.ts) —
  4-test PoC harness:
  1. **VULN** — shows the pre-fix under-count (8% blind spot per 100 tx
     when CU is 3× under-estimated).
  2. **FIX** — same scenario with `adjustForRealizedCost` applied; budget
     converges on actual on-chain spend; drift telemetry reports the gap.
  3. **INTEGRATION** — end-to-end through `keeperSend` with fake timers,
     proves the reconciliation is wired and updates the budget.
  4. **INTEGRATION** — `getTransaction` failure does not halt or propagate.

### Verification

- `pnpm vitest run` — **622 passed / 0 failed** (49 files, 26 skipped).
- `pnpm build` — clean `tsc`.

### Risk

Low. The reconciliation is async / fire-and-forget; the send path is
unchanged on the hot path. Worst-case failure of the reconciliation is
that `realizedCostDriftLamports` stops updating for the affected tx — the
budget keeps its existing (conservative) accounting.

### DIFF
--- a/src/lib/budget.ts
+++ b/src/lib/budget.ts
@@ -51,6 +51,13 @@ export interface BudgetStats {
   haltReason?: string;
   haltKind?: HaltKind;
   config: KeeperBudgetConfig;
+  // M12: cumulative drift between recordTx(estimatedCost) and the actual on-chain
+  // realized cost reconciled via adjustForRealizedCost. Positive = the budget
+  // under-counted (we spent more SOL than we accounted for); negative = the budget
+  // over-counted (we spent less than we accounted for). Exposed so dashboards
+  // can show drift over time and ops can recalibrate priority-fee estimators.
+  realizedCostDriftLamports: number;
+  realizedCostSamples: number;
 }
 
 export type HaltKind =
@@ -137,6 +144,10 @@ export class KeeperBudget {
   private _isHalted = false;
   private _haltKind: HaltKind | undefined;
   private _haltReason: string | undefined;
+
+  // M12: realized-cost reconciliation telemetry.
+  private _realizedCostDriftLamports = 0;
+  private _realizedCostSamples = 0;
 
   constructor(config: Partial<KeeperBudgetConfig> = {}, deps: KeeperBudgetDeps = {}) {
     const envOverrides = parseEnvOverrides(deps.env ?? process.env);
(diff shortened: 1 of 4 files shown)


dcccrypto:
Closed by #216 — rebased onto main (post-M9, post-M10) and squash-merged. M12 budget realized-cost reconciliation is now on main.

PR #228, opened 2026-06-18
fix(liquidation): guard event-driven liquidation dedup

### LINKED ISSUE #218: bug(liquidation): LaserStream Event Path Bypasses Per-Cycle Deduplication and Owner Rate Cap
When `KEEPER_USE_LASERSTREAM=true`, the liquidation service receives real-time account updates and triggers liquidation scans via `onAccount()`. This event-driven path correctly implements a 1-second per-slab debounce that collapses rapid update bursts into a single scan. However, it does not consult `_cycleSeenPositions` (cross-cycle position deduplication) or enforce the `MAX_LIQ_PER_OWNER_PER_CYCLE = 3` per-owner rate cap that the polling path enforces. This creates two distinct failure modes.

***

## Root Cause

The polling path and event-driven path do not share guard state.

**Polling path — guards applied (`liquidation.ts:~626`):**
```typescript
const positionKey = `${candidate.slabAddress}:${candidate.accountIdx}`;
if (this._cycleSeenPositions.has(positionKey)) continue;

const ownerCount = this._cycleOwnerCounts.get(candidate.owner) ?? 0;
if (ownerCount >= LiquidationService.MAX_LIQ_PER_OWNER_PER_CYCLE) continue;
```

**Event-driven path — no guards applied (`liquidation.ts:~733`):**
```typescript
this.scanMarket(market.market).then(async (candidates) => {
  for (const c of candidates) {
    const sig = await this.liquidate(market.market, c.accountIdx, c.scanPriceE6);
    // _cycleSeenPositions not checked
    // _cycleOwnerCounts not checked
    // MAX_LIQ_PER_OWNER_PER_CYCLE not enforced
  }
});
```

***

## Failure Modes

**1. Cross-path duplicate transaction**

The polling scan identifies position `(slabAddress:accountIdx)` as undercollateralised and submits a `LiquidateAtOracle` transaction, recording the position in `_cycleSeenPositions`. Before that transaction lands on-chain, an account update event fires `onAccount()` on the same slab. The event path scans the same slab, finds the same candidate (still appearing undercollateralised because the polling transaction has not settled), and submits a second `LiquidateAtOracle` transaction. One transaction will fail on-chain, but both incur Jito tip and priority fee costs.

**2. Per-owner rate cap bypass**

An owner with five undercollateralised sub-accounts will have all five liquidated simultaneously via the event path. The polling path would cap this at `MAX_LIQ_PER_OWNER_PER_CYCLE = 3` per cycle.

***

## Implementation Plan

- [ ] Apply `_cycleSeenPositions` and `_cycleOwnerCounts` checks inside the `onAccount()` event handler before calling `liquidate()`
- [ ] Decide whether the event path should share the polling cycle's dedup set directly, or maintain a separate per-debounce-window set — document the chosen semantics in a comment
- [ ] Add a unit test that wires both paths against the same mock slab and asserts that a single `liquidate()` call is made per candidate per debounce window

**Suggested guard insertion point:**
```typescript
// In onAccount(), before calling liquidate():
const positionKey = `${market.slabAddress}:${c.accountIdx}`;
if (this._cycleSeenPositions.has(positionKey)) continue;
this._cycleSeenPositions.add(positionKey);

const ownerCount = this._cycleOwnerCounts.get(c.owner) ?? 0;
if (ownerCount >= LiquidationService.MAX_LIQ_PER_OWNER_PER_CYCLE) continue;
this._cycleOwnerCounts.set(c.owner, ownerCount + 1);
```
***

## References

- CWE-799: https://cwe.mitre.org/data/definitions/799.html

Bayyan16:
## Summary

Fixes #218.

This PR applies the same per-cycle liquidation guard to the LaserStream event-driven liquidation path that already protects the polling path.

Before this change, event-driven liquidation candidates could bypass:

- duplicate position protection within the same cycle
- per-owner liquidation cap within the same cycle

## Changes

- Adds a shared `_reserveLiquidationCandidate()` helper.
- Reuses the same guard state for both polling and event-driven liquidation paths:
  - `_cycleSeenPositions`
  - `_cycleOwnerCounts`
  - `MAX_LIQ_PER_OWNER_PER_CYCLE`
- Adds regression coverage for:
  - event path does not re-liquidate a candidate already reserved by polling
  - event path enforces `MAX_LIQ_PER_OWNER_PER_CYCLE`

## Testing

Passed:

```bash
git diff --check
pnpm test tests/services/liquidation-event-dedup.test.ts
```

Result:

```txt
Test Files  1 passed
Tests       2 passed
```

Build note:

```bash
pnpm build
```

is currently blocked by the same existing upstream SDK/local build mismatch on a clean `upstream/main` checkout. The baseline build fails before this PR's changes with missing `@percolatorct/sdk` exports such as `encodePermissionlessCrank`, `CrankAction`, `isV17Account`, and `parsePortfolioV17`.

This PR does not attempt to fix that SDK/keeper mismatch to keep the scope focused on issue #218.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified liquidation target deduplication so the same position isn’t liquidated multiple times within a cycle, across both polling and event-driven execution.
  * Applied consistent per-owner per-cycle liquidation rate limiting for both execution methods.

* **Tests**
  * Added a regression test suite covering event-driven liquidation dedup and per-cycle rate caps.
  * Updated SDK smoke tests to assert runtime failures for v17-incompatible encoders.

* **Chores**
  * Updated CI to clone and build an external SDK dependency before running the main build/test steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,12 @@ jobs:
           node-version: 22
           cache: pnpm
       - name: Checkout linked SDK dependency
-        run: git clone --depth 1 https://github.com/dcccrypto/percolator-sdk.git ../../percolator-sdk
+        run: |
+          mkdir -p ../../percolator-sdk
+          git -C ../../percolator-sdk init
+          git -C ../../percolator-sdk remote add origin https://github.com/dcccrypto/percolator-sdk.git
+          git -C ../../percolator-sdk fetch --depth 1 origin 5b140ae20dbb2e322da8687c7338f7d53821ee81
+          git -C ../../percolator-sdk checkout --detach FETCH_HEAD
       - name: Build linked SDK dependency
         working-directory: ../../percolator-sdk
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,13 @@ jobs:
         with:
           node-version: 22
           cache: pnpm
+      - name: Checkout linked SDK dependency
+        run: git clone --depth 1 https://github.com/dcccrypto/percolator-sdk.git ../../percolator-sdk
+      - name: Build linked SDK dependency
+        working-directory: ../../percolator-sdk
+        run: |
+          pnpm install --frozen-lockfile
+          pnpm build
       - run: pnpm install --frozen-lockfile
       - run: pnpm build
       - run: pnpm test
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -398,6 +398,46 @@ export class LiquidationService {
   }
 
   /**
+   * Shared liquidation target guard for both polling and LaserStream event paths.
+   *
+   * Semantics: the event-driven path intentionally shares the polling cycle guard
+   * state. A target reserved by polling cannot be re-submitted by an event burst
+   * before the next polling cycle clears the guard state. Polling remains active
+   * as the slow-path safety net and resets the window at the start of each cycle.
+   */
+  private _reserveLiquidationCandidate(
+    candidate: LiquidationCandidate,
+    source: "polling" | "event",
+  ): boolean {
+    const positionKey = candidate.v17PortfolioPubkey
+      ? `${candidate.slabAddress}:${candidate.v17PortfolioPubkey.toBase58()}`
+      : `${candidate.slabAddress}:${candidate.accountIdx}`;
+
+    if (this._cycleSeenPositions.has(positionKey)) {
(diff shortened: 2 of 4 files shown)


dcccrypto:
Thanks — this is the same finding as #218, which was just fixed in main via #232 (a shared `gatedLiquidate()` gate applies the per-cycle dedup + MAX_LIQ_PER_OWNER_PER_CYCLE cap on BOTH the polling and LaserStream event paths). Superseded; closing.

PR #242, opened 2026-06-18
Retry v17 keeper portfolio provisioning on rediscovery

### LINKED ISSUE #238: High: v17 markets can remain uncranked forever after transient keeper-portfolio provisioning failure
## Summary

v17 markets require the keeper to have a program-owned keeper portfolio before `PermissionlessCrank` can run. The discovery path provisions that portfolio only when a market is first inserted into `this.markets`. If that async provisioning attempt fails or returns `null`, the market remains in memory with `keeperPortfolio: null`.

On future discovery cycles the market goes through the existing-market branch, which updates metadata and failure counters but never retries `provisionKeeperPortfolio()`. `crankMarket()` and `crankAll()` then skip the market because `keeperPortfolio` is still null.

The comment says provisioning failure will be retried on next discovery, but the code does not do that.

## Impact

A single transient RPC/rent/broadcast race during first discovery can permanently strand a v17 market in this keeper process. The market will be discovered and retained, but it will not be cranked until process restart or manual intervention. That affects funding/refresh/liquidation liveness for the market.

Severity: High operational / liveness bug.

## Code evidence

New markets are inserted with `keeperPortfolio: null`:

```ts
if (!this.markets.has(key)) {
  this.markets.set(key, {
    ...,
    keeperPortfolio: null,
  });
```

Provisioning is attempted only in this new-market branch:

```ts
(async () => {
  const portfolio = await provisionKeeperPortfolio(...);
  const state = this.markets.get(key);
  if (state) {
    state.keeperPortfolio = portfolio;
  }
})();
```

The comment promises retry on next discovery:

```ts
// If provisioning fails (...), the market will be skipped this cycle and retried on next discovery.
```

But the existing-market branch only updates state and never calls `provisionKeeperPortfolio()` when `state.keeperPortfolio` remains null.

Both crank paths skip null portfolios:

```ts
if (!state.keeperPortfolio) {
  logger.debug("Skipping crank — keeper portfolio not provisioned yet", { slabAddress });
  return false;
}
```

and:

```ts
if (!state.keeperPortfolio) {
  skippedNoPortfolio++;
  continue;
}
```

## Repro PoC

This state-machine reproducer mirrors the keeper control flow:

```bash
node <<'NODE'
let markets = new Map();
const key = 'market-A';
let provisionAttempts = 0;

function discoverMarket(provisionSucceeds) {
  if (!markets.has(key)) {
    markets.set(key, { keeperPortfolio: null, market: key });
    provisionAttempts++;
    if (provisionSucceeds) {
      markets.get(key).keeperPortfolio = 'portfolio-A';
    }
  } else {
    const state = markets.get(key);
    state.market = key; // existing branch refreshes metadata only
    // BUG: no retry when state.keeperPortfolio is still null
  }
}

function crankAllWouldSkip() {
  return markets.get(key).keeperPortfolio === null;
}

// First discovery sees the market, but provisioning has a transient failure.
discoverMarket(false);

// Next discovery cycle sees the same market again; provisioning could now succeed,
// but the existing-market branch never retries it.
discoverMarket(true);

console.log({
  provisionAttempts,
  keeperPortfolio: markets.get(key).keeperPortfolio,
  skippedByCrankAll: crankAllWouldSkip(),
});

if (provisionAttempts === 1 && crankAllWouldSkip()) {
  console.error('POC: market remains discovered but permanently skipped after one transient provisioning failure');
  process.exit(1);
}
NODE
```

Output:

```text
{ provisionAttempts: 1, keeperPortfolio: null, skippedByCrankAll: true }
POC: market remains discovered but permanently skipped after one transient provisioning failure
```

## Suggested fix

Retry provisioning for existing v17 markets whose `state.keeperPortfolio` is still null. A safe shape is:

1. Extract the current async provisioning block into a helper like `ensureKeeperPortfolio(key, market, state)`.
2. Call it when inserting a new v17 market.
3. Also call it from the existing-market branch when `state.keeperPortfolio === null`, with the existing in-flight guard/backoff behavior so rediscovery does not spam transactions.
4. Add a regression where first provisioning returns `null`, second discovery retries, and `crankAll()` no longer increments `skippedNoPortfolio` after the retry succeeds.

RehanNek:
## Summary

Fixes #238.

Extracts the v17 keeper-portfolio provisioning block into `ensureKeeperPortfolio()` and calls it:

- when a market is first discovered
- again on rediscovery whenever the tracked state still has `keeperPortfolio === null`

This makes the existing comment true: transient provisioning failures are retried on subsequent discovery cycles instead of leaving the market permanently skipped until process restart.

## Tests

```bash
pnpm run build
```

### DIFF
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -838,6 +838,42 @@ export class CrankService {
   }
 
   /**
+   * v17 markets need a keeper-owned portfolio before PermissionlessCrank can run.
+   * Provisioning can fail transiently, so discovery calls this on first insert and
+   * again on rediscovery while keeperPortfolio is still null.
+   */
+  private ensureKeeperPortfolio(key: string, market: DiscoveredMarket): void {
+    void (async () => {
+      try {
+        const connection = getConnection();
+        const keypair = this._keypair;
+        const portfolio = await provisionKeeperPortfolio(
+          connection,
+          market.programId,
+          market.slabAddress,
+          keypair.publicKey,
+          keypair,
+        );
+        const state = this.markets.get(key);
+        if (state) {
+          state.keeperPortfolio = portfolio;
+          if (portfolio) {
+            logger.info("Keeper portfolio provisioned", {
+              market: key.slice(0, 8),
+              portfolio: portfolio.toBase58().slice(0, 8),
+            });
+          }
+        }
+      } catch (provErr) {
+        logger.debug("Portfolio provisioning deferred", {
+          market: key.slice(0, 8),
+          error: provErr instanceof Error ? provErr.message : String(provErr),
+        });
+      }
+    })();
+  }
+
+  /**
    * PERC-1650: Per-program 429 retry backoff for discoverMarkets calls.
    * Escalating delays: 3s → 9s → 27s → 81s before giving up.
    * Applied at the program level (outer loop).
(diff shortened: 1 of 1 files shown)


dcccrypto:
Superseded by #250 (merged) — your fix rebased onto current main; the retry is bounded by the existing provisionKeeperPortfolio._inflight dedup. Authorship preserved, finding credit stands. The #238-required regression test is filed as a follow-up issue (the fire-and-forget async-private path needs a careful harness; didn't want to block the stranded-markets fix). Thanks @RehanNek!

PR #258, opened 2026-06-19
fix: refresh v17 liquidation scan price

### LINKED ISSUE #257: Medium: v17 liquidation scan uses cached price instead of freshly fetched wrapper price
## Summary

`LiquidationService.scanMarket()` fetches fresh v17 market account data, but the v17 branch currently computes the scan price from the cached `market.config` object created during discovery:

```ts
const price = market.config.lastEffectivePriceE6 ?? market.config.authorityPriceE6 ?? 0n;
```

For v17 markets this can use an old discovery-time price even though `scanMarket()` already has the fresh wrapper account bytes in hand. It also treats `0n` as a valid `lastEffectivePriceE6` value because `??` only falls back on `null`/`undefined`, so a cached zero mark can suppress a fresh nonzero authority target.

## Impact

A v17 AUTH_MARK market can be skipped during liquidation scanning even when the freshly fetched wrapper account contains a valid authority price. More generally, v17 scan candidacy can lag behind the actual market wrapper price until rediscovery refreshes `market.config`, causing missed liquidation opportunities or stale candidacy decisions.

Severity: Medium. This is a liveness/correctness issue in the keeper scanner; it does not require malformed account data.

## Proof of Finding

I added a regression test that sets up:

- a v17 market account (`isV17Account(data) === true`),
- cached `market.config.lastEffectivePriceE6 = 0n`,
- a freshly fetched v17 wrapper config with `oracleMode = AUTH_MARK`, `oracleTargetPriceE6 = 2_000_000n`, and a fresh publish timestamp,
- one undercollateralized v17 portfolio returned by `getProgramAccounts`.

Expected behavior: the scanner parses the fresh wrapper config and returns one candidate with `scanPriceE6 = 2_000_000n`.

Current behavior before the fix: the scanner uses cached `0n`, `scanV17Portfolios()` returns no candidates, and the position is missed.

Repro command:

```bash
pnpm exec vitest run tests/services/liquidation.test.ts -t "fresh v17 AUTH_MARK price"
```

With the fix, the focused repro and full file pass:

```bash
pnpm exec vitest run tests/services/liquidation.test.ts
pnpm exec tsc --noEmit
```

## Suggested Fix

In the v17 branch of `scanMarket()`:

1. parse the freshly fetched wrapper bytes with `parseWrapperConfigV17(data)`,
2. resolve the scan price from the fresh v17 config,
3. for AUTH_MARK, prefer a fresh `oracleTargetPriceE6`; otherwise use the EWMA mark,
4. pass that fresh price into `scanV17Portfolios()` and store it in `scanPriceE6`.

## Non-duplicate check

This is separate from #245. #245 covers the v17 pre-submit liquidation recheck/drift guard using stale scan-time price after a candidate has already been found. This issue is earlier in the flow: `scanMarket()` can fail to find the v17 candidate in the first place because it uses cached discovery-time price instead of the fresh wrapper price it already fetched.

RehanNek:
## Summary

Fixes #257.

`scanMarket()` now derives the v17 liquidation scan price from the freshly fetched wrapper account data instead of stale discovery-time `market.config`. For AUTH_MARK v17 markets it uses a fresh `oracleTargetPriceE6` when the wrapper timestamp is within staleness bounds, otherwise it falls back to the EWMA mark.

## Proof / Tests

Added a regression test proving that a cached zero `market.config.lastEffectivePriceE6` no longer suppresses a fresh v17 AUTH_MARK authority price.

```bash
pnpm exec vitest run tests/services/liquidation.test.ts
pnpm exec tsc --noEmit
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated liquidation pricing for v17 accounts to fetch fresh oracle prices at scan time with a 60-second freshness check, automatically falling back to alternative pricing when oracle data is stale.

* **Tests**
  * Added test coverage for v17 oracle pricing behavior in liquidation scans.

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -12,6 +12,7 @@ import {
   CrankAction,
   // v17 portfolio scanning (DESYNC-3 / DESYNC-4 fixes)
   isV17Account,
+  parseWrapperConfigV17,
   parsePortfolioV17,
   type DiscoveredMarket,
 } from "@percolatorct/sdk";
@@ -318,6 +319,26 @@ function resolveMarketPrice(
   return { price: cfg.lastEffectivePriceE6, stale: true };
 }
 
+function resolveV17ScanPrice(
+  cfg: ReturnType<typeof parseWrapperConfigV17>,
+  nowSec: bigint,
+): bigint {
+  // Program enum (v16_program.rs): MANUAL=0, HYBRID_AFTER_HOURS=1,
+  // EWMA_MARK=2, AUTH_MARK=3. AUTH_MARK is the only v17 mode whose fresh
+  // authority target should beat the EWMA fallback during liquidation scans.
+  if (cfg.oracleMode === 3) {
+    const maxStalenessSecs = cfg.maxStalenessSecs > 0n ? cfg.maxStalenessSecs : 60n;
+    const priceAge = cfg.oracleTargetPublishTime > 0n
+      ? nowSec - cfg.oracleTargetPublishTime
+      : nowSec;
+    if (cfg.oracleTargetPriceE6 > 0n && priceAge <= maxStalenessSecs) {
+      return cfg.oracleTargetPriceE6;
+    }
+  }
+
+  return cfg.markEwmaE6;
+}
+
 interface LiquidationCandidate {
   slabAddress: string;
   accountIdx: number;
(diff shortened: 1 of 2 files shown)


dcccrypto:
Superseded by the merged #271. The substantive #257 fix — resolving a FRESH wrapper price (AUTH_MARK mode, with a staleness check, falling back to markEwmaE6) instead of the cached scan-time price — already landed in #271 as `resolveV17WrapperPrice` (liquidation.ts), which is invoked on the v17 scan path. This PR only renames that function to `resolveV17ScanPrice` and adds an explanatory comment; the behavior is already present. Closing as redundant — credit @RehanNek for the #257 finding (fixed via #271).

PR #280, opened 2026-06-20
fix(keeper): close stop()/connect() race in AccountLoader

Morenikeoa:
## Bug

`AccountLoader.connect()` (`src/lib/account-loader.ts`) awaits `_snapshotKnownAccounts()` then `adapter.start(...)`. If `stop()` is called while either await is in flight, `stop()` sets `running=false` and calls `adapter.stop()` — but at that point the adapter's internal handle hasn't been assigned yet (`LaserStreamAdapter.handle` is set only once `subscribe()` resolves), so the call is a no-op. When `connect()`'s await later resolves, it unconditionally sets `connected = true` with no check of `running`, leaving a live, otherwise-uncancelled subscription running after the loader was told to stop. `enqueue()` also has no `running` guard at all, unlike `onStreamError()` which does.

## Impact

This is no longer a latent path — `index.ts`'s `stopAllServices()` (called on HA demotion) awaits `accountLoader.stop()`, while `startAllServices()` elsewhere may still have `accountLoader.start()` in flight (e.g. a leader-lock flap that promotes then immediately demotes a node). When that races, the loader reports `connected: true` after being told to stop, the underlying stream subscription is never cancelled, and it keeps mutating the shared `AccountCache` and firing registered listeners indefinitely — risking stale-looking-fresh cache reads by `CrankService`/`LiquidationService` even though the node believes its account stream was torn down.

## Fix

`connect()` re-checks `this.running` immediately after each await, before committing to connected state:
- After `_snapshotKnownAccounts()` resolves: bail out before even starting the stream subscription if `stop()` already ran.
- After `adapter.start()` resolves: if `stop()` raced in, call `adapter.stop()` again — this time against a real, live handle — and return without setting `connected = true`.

`enqueue()` gets a `running` guard matching the existing convention in `onStreamError()`, as defense-in-depth against any update delivered in the brief window between cancellation being requested and the underlying transport actually tearing down.

## Test results

New tests added to `tests/lib/account-loader.test.ts` (a `DeferredStartAdapter` test double whose `start()` doesn't resolve until the test releases it, to put `connect()` in a controllable mid-await state; written first, confirmed failing against the unfixed code, then confirmed passing after the fix):

```
 Test Files  1 passed (1)
      Tests  27 passed | 1 skipped (28)
```

Full suite, post-fix:

```
 Test Files  83 passed | 2 skipped (85)
      Tests  868 passed | 33 skipped (901)
```

`pnpm build` passes cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved race conditions in account loading operations that could occur with simultaneous stop and connect actions.
  * Ensured account updates are properly discarded after the loader stops, preventing stale data processing.
  * Improved shutdown sequence reliability with additional safeguards to prevent reconnection after being stopped.

### DIFF
--- a/src/lib/account-loader.ts
+++ b/src/lib/account-loader.ts
@@ -261,6 +261,11 @@ export class AccountLoader {
       // cache.set() calls and overwrite newer slot data with older RPC data.
       await this._snapshotKnownAccounts();
 
+      // H-5: stop() may have run while the snapshot RPC above was in flight.
+      // Don't bother starting a stream subscription for a loader that's
+      // already been told to stop.
+      if (!this.running) return;
+
       await this.adapter.start(
         this.opts,
         (update) => this.enqueue(update),
@@ -270,6 +275,18 @@ export class AccountLoader {
         },
         (err) => this.onStreamError(err),
       );
+
+      // H-5: stop() may instead have run while adapter.start() itself was in
+      // flight. At that point stop()'s call to adapter.stop() found no handle
+      // yet assigned (e.g. LaserStreamAdapter.handle is set only once
+      // subscribe() resolves) and was a no-op -- the subscription that just
+      // came up here is live and otherwise uncancelled. Tear it down now
+      // instead of marking the loader connected.
+      if (!this.running) {
+        this.adapter.stop();
+        return;
+      }
+
       this.connected = true;
       this.backoff.reset();
       logger.info("AccountLoader: stream connected", {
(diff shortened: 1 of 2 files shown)


dcccrypto:
Closing as duplicate — #281 is a strict superset (it ships this same H-5 AccountLoader stop()/connect() race fix verbatim, plus the H-7 mainnet Discord-webhook fail-loud). Merging both would double-apply the H-5 change. Credit @Morenikeoa — your fix lands via #281.

PR #287, opened 2026-06-20
fix(keeper): pass expectedOwner to fetchSlab to enforce on-chain owner check

Morenikeoa:
## Bug

`fetchSlab(connection, slabPubkey, expectedOwner?)` (from `@percolatorct/sdk`) only checks that the fetched account is owned by the Percolator program when `expectedOwner` is passed:

```ts
if (expectedOwner && !info.owner.equals(expectedOwner)) {
  throw new Error(`fetchSlab: account ${slabPubkey} is owned by ${info.owner} but expected ${expectedOwner}`);
}
```

Both real call sites in this repo omit it:
- `liquidation.ts`'s `fetchSlabWithRetry` (used by `scanMarket`, and both the v17 and v12.x pre-submit rechecks in `liquidate`)
- `monitor.ts`'s conservation-invariant check

Without `expectedOwner`, the only validation left is the caller's later magic-byte parsing (`parseHeader`/`isV17Account`). A magic-byte match at the right offsets is not equivalent to an owner check — a misrouted or crafted account (e.g. from a compromised/buggy RPC response, or any future code path that derives a slab pubkey from less-trusted input) with the right header layout would be silently accepted as a legitimate market/slab account, feeding liquidation decisions, margin-ratio checks, and the crank's conservation-invariant monitor from data the Percolator program never actually wrote.

## Fix

Pass the market's `programId` (already in scope as `DiscoveredMarket.programId` at every call site) through as `expectedOwner`:
- `liquidation.ts`: `fetchSlabWithRetry` now takes an `expectedOwner: PublicKey` parameter, threaded through to all 3 call sites (`scanMarket`, v17 recheck, v12.x recheck).
- `monitor.ts`: `fetchSlab(conn, crankState.market.slabAddress, crankState.market.programId)`.

Pure plumbing — no new RPC calls, no behavior change for the happy path (where the account genuinely is owned by the program).

## Test plan

- New assertions confirming `fetchSlab` is called with the market's `programId` as the third argument, at all 4 affected call sites (`tests/services/liquidation.test.ts` x3, `tests/services/monitor.test.ts` x1).
- Confirmed all 4 fail against the pre-fix code (`git stash` the source change, re-run).
- `pnpm build` — clean.
- `pnpm test` — full suite green, no regressions (866 passed / 33 skipped).

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -54,6 +54,7 @@ function withTimeout<T>(promise: Promise<T>, ms: number, label: string): Promise
  */
 async function fetchSlabWithRetry(
   slabPubkey: PublicKey,
+  expectedOwner: PublicKey,
   maxRetries = 3,
 ): Promise<Uint8Array> {
   let lastErr: unknown;
@@ -62,7 +63,7 @@ async function fetchSlabWithRetry(
     try {
       await acquireToken();
       return await withTimeout(
-        fetchSlab(conn, slabPubkey),
+        fetchSlab(conn, slabPubkey, expectedOwner),
         RPC_TIMEOUT_MS,
         `fetchSlab(${slabPubkey.toBase58()})`,
       );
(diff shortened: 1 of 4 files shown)


dcccrypto:
Rebased + merged as #289 (your fork branch conflicted with merged #282 in liquidation.ts; kept the H-8 handling + threaded programId). Full credit @Morenikeoa — closing as merged-equivalent.

Issue #326, opened 2026-06-22
Security: SDK dependency pinned to unverified git SHA in production while CI tests npm release


v1ktorrr0x:
## Bug

`package.json` installs `@percolatorct/sdk` from a raw git commit SHA rather than a signed npm release:

```json
"@percolatorct/sdk": "git+https://github.com/dcccrypto/percolator-sdk.git#13f3012918003d995ee8a3793d8817b8c19f9762"
```

Meanwhile, `sdk-smoke.yml` CI tests against the published npm version (`SDK_VERSION: "2.0.5"`). Production builds and CI are testing fundamentally different code paths.

## Impact

- **Supply chain risk:** A force-push or history rewrite at the pinned SHA on `percolator-sdk` silently changes what production installs — no npm provenance, no integrity hash, no audit trail.
- **CI/prod drift:** CI validates npm `2.0.5`; production runs whatever is at the git SHA. A regression present in the git SHA but fixed before the npm publish would ship to production undetected.
- **No lockfile protection:** `pnpm-lock.yaml` records the git SHA, not a content hash of the resolved tarball, so `pnpm install --frozen-lockfile` provides weaker guarantees than a registry dep.

## Fix

1. Verify that npm `2.0.5` corresponds to the pinned git SHA.
2. Replace the git dep with the registry version in `package.json`:
   ```json
   "@percolatorct/sdk": "2.0.5"
   ```
3. Regenerate the lockfile and commit. CI and production will now test and run identical code.

## Source

- `package.json` — `dependencies["@percolatorct/sdk"]`
- `.github/workflows/sdk-smoke.yml` — `env.SDK_VERSION: "2.0.5"`
- References: [SLSA Supply Chain Threats](https://slsa.dev/spec/v1.0/threats), [npm provenance docs](https://docs.npmjs.com/generating-provenance-statements)

dcccrypto:
[...]

Issue #324, opened 2026-06-22
Bug: /register HTTP handler skips on-chain ownership pre-validation, causing unnecessary RPC calls on invalid input


v1ktorrr0x:
## Bug

The `POST /register` HTTP handler validates only the base58 format of `slabAddress` before calling `crankService.registerMarket()`. All on-chain owner validation is deferred to `registerMarket()` inside `crank.ts`, which must make a live RPC call (`getAccountInfo`) to fetch the account before it can check whether the owner is in `config.allProgramIds`.

```typescript
// src/index.ts — only validation before RPC call:
const base58Re = /^[1-9A-HJ-NP-Za-km-z]{32,44}$/;
if (!base58Re.test(slabAddress)) { ... } // format only
// → immediately calls crankService.registerMarket(slabAddress, mainnetCA)
```

`config.allProgramIds` is available at module scope in `index.ts` at zero cost.

## Impact

- Every authenticated request with a valid base58 address (but unknown program owner) burns one `getAccountInfo` RPC call before being rejected.
- A rate-limited attacker (5 failures per 60s per IP) can cause 5 RPC calls per window per IP purely from the HTTP layer, before any program ownership is checked. Under a distributed attack this accumulates RPC spend.
- Adds latency to the rejection path for legitimate misconfiguration errors.

## Fix

Add a lightweight program allow-list pre-check directly in the HTTP handler, before calling `registerMarket()`:

```typescript
// After base58 format check, before calling registerMarket:
const knownProgramIds = new Set(config.allProgramIds);
// Pre-check is best-effort (we don't have owner yet without an RPC call),
// but we can at least reject obviously wrong addresses that match known-bad patterns.
// The full ownership check in registerMarket() remains as the authoritative gate.
```

Alternatively, expose the program allow-list check as a static helper in `crank.ts` and call it synchronously before the RPC fetch, so the HTTP handler can fail-fast with a `400` before any network I/O occurs.

## Source

- `src/index.ts` → `POST /register` handler, base58 validation block
- `src/services/crank.ts` → `registerMarket()` → `getAccountInfo` → `knownIds.has(info.owner.toBase58())`

dcccrypto:
[...]

PR #327, opened 2026-06-22
Fix security audit issues

### LINKED ISSUE #314: Security: Shell injection via unguarded expression interpolation in sdk-smoke.yml
## Bug

`github.event.client_payload.version` (from a `repository_dispatch` event) and `inputs.sdk_version` (from `workflow_dispatch`) are interpolated directly into `run:` shell steps using `${{ ... }}` syntax — without routing through an `env:` block first. Any party that can send a `repository_dispatch` to this repo can inject arbitrary shell commands into the CI runner.

Four affected steps in `.github/workflows/sdk-smoke.yml`:

```yaml
# Step: Resolve SDK version
VERSION="${{ github.event.client_payload.version }}"

# Step: Override SDK dep
pnpm add "@percolatorct/sdk@${{ steps.sdk_ver.outputs.version }}"

# Step: Verify installed SDK version
if [ "$INSTALLED" != "${{ steps.sdk_ver.outputs.version }}" ]; then

# Step: Report result
echo "@percolatorct/sdk@${{ steps.sdk_ver.outputs.version }} smoke PASSED"
```

`steps.sdk_ver.outputs.version` is itself populated from the raw `client_payload.version` value, so all four steps are part of the same injection chain.

## Impact

An attacker with the ability to trigger a `repository_dispatch` (any collaborator, or an exploited downstream service holding a PAT with `repo` scope) can execute arbitrary shell commands in the CI runner — including exfiltrating secrets (e.g. `NPM_TOKEN`, `GITHUB_TOKEN`), modifying build artifacts, or pivoting to npm publish abuse.

CVSS v3.1 estimate: **8.8 High** (AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:N)

## Fix

Move all context expressions into `env:` blocks and reference them as environment variables in the shell. GitHub Actions evaluates `env:` values in a safe context and never passes them through the shell parser as code.

```yaml
- name: Resolve SDK version
  id: sdk_ver
  env:
    CLIENT_PAYLOAD_VERSION: ${{ github.event.client_payload.version }}
    INPUT_VERSION: ${{ inputs.sdk_version }}
    PINNED_VERSION: ${{ env.SDK_VERSION }}
  run: |
    VERSION="${CLIENT_PAYLOAD_VERSION:-${INPUT_VERSION:-$PINNED_VERSION}}"
    echo "version=$VERSION" >> "$GITHUB_OUTPUT"
    echo "Testing @percolatorct/sdk@$VERSION"

- name: Override SDK dep with published npm version
  env:
    SDK_VER: ${{ steps.sdk_ver.outputs.version }}
  run: pnpm add "@percolatorct/sdk@${SDK_VER}" --no-frozen-lockfile
```

Apply the same `env:` pattern to the Verify and Report steps.

## Source

- File: `.github/workflows/sdk-smoke.yml`
- Steps: `Resolve SDK version`, `Override SDK dep with published npm version`, `Verify installed SDK version`, `Report result`
- References: [GitHub Security Lab — Expression injection](https://securitylab.github.com/research/github-actions-untrusted-input/), [CWE-78](https://cwe.mitre.org/data/definitions/78.html)

### LINKED ISSUE #321: Security: /health, /pause-status, and /shadow/report expose sensitive operational data without authentication
## Bug

`GET /health`, `GET /pause-status`, and `GET /shadow/report` are served by the health HTTP server with **zero authentication**. Any client that can reach the bind address receives a full operational snapshot.

`/health` response includes:
- Keeper wallet SOL balance and low-balance flag
- HA leader/standby role (`leaderLock.role()`)
- Budget circuit-breaker state: `halted`, `haltKind`, `cycleSpend`, `hourSpend`, `daySpend`
- Stale oracle market addresses and pause state
- Liquidation scanner stats: `scanCount`, `liquidationCount`, `lastScanTime`
- Full sender metrics snapshot

`/pause-status` returns paused market addresses.

`/shadow/report` (when `SHADOW_HARNESS_ENABLED=true`) returns the internal decision log.

The server binds to `127.0.0.1` by default, but `KEEPER_HEALTH_BIND_ADDR=0.0.0.0` — a common pattern on Railway, Kubernetes, or Docker deployments that need liveness probes — exposes all three endpoints publicly with no protection.

## Impact

- Wallet balance and address leaks inform timing attacks (e.g. sandwich the keeper when balance is low).
- HA role exposure allows an attacker to identify and target the active leader specifically.
- Budget circuit-breaker state reveals whether the keeper is halted and why, informing exploit timing.
- Decision log (`/shadow/report`) leaks internal crank logic and oracle price decisions.

## Fix

Add shared-secret authentication to all non-trivial health endpoints when `KEEPER_HEALTH_BIND_ADDR` is not `127.0.0.1`:

```typescript
function requireAuth(req: http.IncomingMessage, res: http.ServerResponse): boolean {
  if (healthBindAddr === "127.0.0.1") return true; // loopback: trust network boundary
  const secret = process.env.KEEPER_HEALTH_SECRET ?? "";
  if (!secret) return true; // not configured: fail open with a warning at startup
  const provided = String(req.headers["x-shared-secret"] ?? "");
  // constant-time compare (reuse existing pattern from /register)
  if (!timingSafeCompare(secret, provided)) {
    res.writeHead(401, secureJsonHeaders);
    res.end(JSON.stringify({ error: "Unauthorized" }));
    return false;
  }
  return true;
}
```

Alternatively, strip sensitive fields (`budget`, `senderMetrics`, `wallet`) from the unauthenticated `/health` response and expose them only on a separate authenticated `/health/detail` endpoint.

## Source

- `src/index.ts` → `GET /health`, `GET /pause-status`, `GET /shadow/report` handlers
- No auth check exists on any of these three routes in the current code

### LINKED ISSUE #322: Security: Rate-limit map uses FIFO eviction allowing authenticated brute-force bypass
## Bug

The `registerFailures` map that rate-limits `/register` and `/admin/budget/resume` evicts entries in insertion order (FIFO) when it reaches 10,000 entries:

```typescript
if (registerFailures.size > 10_000) {
  const oldest = registerFailures.keys().next().value;
  if (oldest !== undefined) registerFailures.delete(oldest);
}
```

An attacker can send 10,001 auth-failure requests from unique spoofed IPs to push their own IP out of the map entirely, clearing their 5-failure rate-limit record mid-attack and starting fresh.

`lru-cache` is already a declared dependency (`"lru-cache": "^11.5.0"` in `package.json`) but is not used here.

## Impact

The rate limiter provides no meaningful protection against a determined attacker who controls even a modest botnet or can spoof source IPs. The `/register` secret and `/admin/budget/resume` secret become brute-forceable given enough parallel IPs.

## Fix

Replace the plain `Map` with an `LRUCache` from the already-available dependency so eviction is LRU (least-recently-used), making map flooding ineffective:

```typescript
import { LRUCache } from "lru-cache";

// Replace:
const registerFailures = new Map<string, number[]>();

// With:
const registerFailures = new LRUCache<string, number[]>({ max: 10_000 });
```

The `isRateLimited`, `recordAuthFailure`, and cleanup interval logic require no other changes — `LRUCache` implements the same `get`/`set`/`delete`/`entries` interface.

The periodic cleanup interval (`rateLimitCleanupTimer`) can also be removed since LRU eviction handles memory bounding automatically.

## Source

- `src/index.ts` → `registerFailures` map, `isRateLimited()`, `recordAuthFailure()`
- `package.json` → `dependencies["lru-cache"]` (already present, unused here)

### LINKED ISSUE #323: Security: Single shared secret used for both /register and /admin/budget/resume violates least privilege
## Bug

Both `/register` and `/admin/budget/resume` authenticate using the same environment variable:

```typescript
// /register
const registerSecret = process.env.KEEPER_REGISTER_SECRET ?? "";

// /admin/budget/resume — identical line
const registerSecret = process.env.KEEPER_REGISTER_SECRET ?? "";
```

These endpoints have very different privilege levels. Registering a market is a routine operation performed by the frontend on each new market creation. Resuming a budget circuit-breaker halt is a high-trust administrative action that should only be performed by an operator during an active incident.

## Impact

Any service or person holding `KEEPER_REGISTER_SECRET` — including a compromised frontend, a leaked `.env`, or a junior operator — can also clear a budget circuit-breaker halt. This undermines the safety guarantee of the circuit breaker: the halt may have been set intentionally during a security incident, and an automated or malicious resume would cause the keeper to continue sending transactions into a compromised environment.

## Fix

Introduce a separate `KEEPER_ADMIN_SECRET` environment variable for the `/admin/*` namespace:

```typescript
// /admin/budget/resume
const adminSecret = process.env.KEEPER_ADMIN_SECRET ?? process.env.KEEPER_REGISTER_SECRET ?? "";
// (fallback to KEEPER_REGISTER_SECRET during migration only; remove fallback in next release)
```

Document in `.env.example` that `KEEPER_ADMIN_SECRET` must be a distinct value from `KEEPER_REGISTER_SECRET` and must never be shared with frontend services.

## Source

- `src/index.ts` → `POST /register` handler (line: `const registerSecret = process.env.KEEPER_REGISTER_SECRET`)
- `src/index.ts` → `POST /admin/budget/resume` handler (same variable, same value)

### LINKED ISSUE #325: Bug: KEEPER_HEALTH_BIND_ADDR and STALE_ALERT_COOLDOWN_MS lack startup input validation
## Bug

Two environment variables are consumed without any validation at boot, unlike `KEEPER_HEALTH_PORT` which has an explicit integer range check that throws on misconfiguration.

### 1. `KEEPER_HEALTH_BIND_ADDR`

```typescript
const healthBindAddr = process.env.KEEPER_HEALTH_BIND_ADDR ?? "127.0.0.1";
// No format check. Passed directly to:
healthServer.listen(healthPort, healthBindAddr, () => { ... });
```

An invalid value (e.g. `"127.0.0..1"`, `"locahost"`) causes a confusing `EADDRNOTAVAIL` or `ERR_INVALID_IP_ADDRESS` crash at `listen()` time — not at module load — making it hard to distinguish from a port conflict or OS-level bind error.

### 2. `KEEPER_STALE_ALERT_COOLDOWN_MS`

```typescript
const STALE_ALERT_COOLDOWN_MS = Number(process.env.KEEPER_STALE_ALERT_COOLDOWN_MS ?? 5 * 60_000);
```

- `Number("0")` → `0`: the stale-oracle check fires a Discord critical alert on **every** 60-second interval tick, flooding the channel.
- `Number("abc")` → `NaN`: all cooldown comparisons (`now - last >= NaN`) return `false`, silently disabling all stale oracle alerts.
- `Number("-1")` → `-1`: alerts fire on every tick (same as 0).

## Impact

- Misconfigured `KEEPER_HEALTH_BIND_ADDR` causes a non-obvious crash rather than a clean startup failure, delaying diagnosis.
- A misconfigured or maliciously set `KEEPER_STALE_ALERT_COOLDOWN_MS=0` floods Discord, potentially causing alert fatigue or channel rate-limiting that masks real incidents. `NaN` silently disables oracle staleness alerting entirely.

## Fix

```typescript
// KEEPER_HEALTH_BIND_ADDR
import { isIP } from "node:net";
if (healthBindAddr !== "localhost" && isIP(healthBindAddr) === 0) {
  throw new Error(
    `Invalid KEEPER_HEALTH_BIND_ADDR: "${healthBindAddr}" — must be a valid IPv4/IPv6 address or "localhost"`
  );
}

// KEEPER_STALE_ALERT_COOLDOWN_MS
if (!Number.isFinite(STALE_ALERT_COOLDOWN_MS) || STALE_ALERT_COOLDOWN_MS < 60_000) {
  throw new Error(
    `Invalid KEEPER_STALE_ALERT_COOLDOWN_MS: "${process.env.KEEPER_STALE_ALERT_COOLDOWN_MS}" — must be a finite integer ≥ 60000`
  );
}
```

Both checks should live alongside the existing `KEEPER_HEALTH_PORT` validation block for consistency.

## Source

- `src/index.ts` → `healthBindAddr` declaration
- `src/index.ts` → `STALE_ALERT_COOLDOWN_MS` declaration
- `src/index.ts` → `KEEPER_HEALTH_PORT` validation block (reference implementation for the fix pattern)

v1ktorrr0x:
*   **Fixed CI Shell Injection (#314)**
    Refactored the `sdk-smoke.yml` workflow to pass the client payload version into a secure `env:` block rather than directly interpolating `${{ github.event.client_payload.version }}` into the `run:` script. This prevents potential attacker-controlled shell command execution via `repository_dispatch` events.
    
*   **Secured Monitoring Endpoints (#321)**
    Added an authentication barrier for the `/health`, `/pause-status`, and `/shadow/report` HTTP endpoints when the server is bound to a remote address (anything other than `127.0.0.1`, `localhost`, or `::1`). Remote access now requires the `x-shared-secret` header, which is validated using a constant-time `timingSafeEqual` comparison against `KEEPER_REGISTER_SECRET`.
    
*   **Plugged Rate-Limit Map Memory Leak (#322)**
    Replaced the unbounded standard `Map` previously used for `/register` rate limiting with a capped `LRUCache`. This protects the service against memory exhaustion (OOM) vectors from an infinite buildup of attacker IPs.
    
*   **Separated Administrative Secrets (#323)**
    Updated the `/admin/budget/resume` circuit-breaker endpoint to read from a newly isolated `KEEPER_ADMIN_SECRET` environment variable, ensuring that standard registration workflows do not share privileges with administrative capabilities.
    
*   **Hardened Environment Configuration (#325)**
    Implemented fail-fast startup validations for critical environment variables:
    *   `KEEPER_HEALTH_BIND_ADDR`: Strictly validated using `node:net` to ensure it is a legitimate IP address or `localhost`.
    *   `KEEPER_STALE_ALERT_COOLDOWN_MS`: Validated to be a finite number `>= 60000` to prevent rate-limit flooding of the Discord alerting API.
    
*   **Cleaned Docker Image Layers**
    Removed the `store-dir=.pnpm-store` directive from `.npmrc` to correct a configuration bug that resulted in bloated Docker layers through duplicated content-addressable storage.

### 🔗 Linked Issues
* Fixes #314
* Fixes #321
* Fixes #322
* Fixes #323
* Fixes #325

### DIFF
--- a/.github/workflows/sdk-smoke.yml
+++ b/.github/workflows/sdk-smoke.yml
@@ -58,13 +58,16 @@ jobs:
       # input > pinned default.
       - name: Resolve SDK version
         id: sdk_ver
+        env:
+          PAYLOAD_VER: ${{ github.event.client_payload.version }}
+          INPUT_VER: ${{ inputs.sdk_version }}
         run: |
-          VERSION="${{ github.event.client_payload.version }}"
+          VERSION="$PAYLOAD_VER"
           if [ -z "$VERSION" ]; then
-            VERSION="${{ inputs.sdk_version }}"
+            VERSION="$INPUT_VER"
           fi
           if [ -z "$VERSION" ]; then
-            VERSION="${{ env.SDK_VERSION }}"
+            VERSION="${SDK_VERSION}"
           fi
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
           echo "Testing @percolatorct/sdk@$VERSION"
@@ -77,16 +80,20 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Override SDK dep with published npm version
+        env:
+          TARGET_VER: ${{ steps.sdk_ver.outputs.version }}
         run: |
-          pnpm add "@percolatorct/sdk@${{ steps.sdk_ver.outputs.version }}" \
+          pnpm add "@percolatorct/sdk@${TARGET_VER}" \
             --no-frozen-lockfile
 
       - name: Verify installed SDK version
+        env:
+          TARGET_VER: ${{ steps.sdk_ver.outputs.version }}
         run: |
           INSTALLED=$(node -e "console.log(require('./node_modules/@percolatorct/sdk/package.json').version)")
           echo "Installed: $INSTALLED"
-          if [ "$INSTALLED" != "${{ steps.sdk_ver.outputs.version }}" ]; then
-            echo "Version mismatch: expected ${{ steps.sdk_ver.outputs.version }}, got $INSTALLED"
+          if [ "$INSTALLED" != "$TARGET_VER" ]; then
+            echo "Version mismatch: expected $TARGET_VER, got $INSTALLED"
             exit 1
           fi
 
@@ -95,10 +102,13 @@ jobs:
 
       - name: Report result
         if: always()
+        env:
+          TARGET_VER: ${{ steps.sdk_ver.outputs.version }}
+          JOB_STATUS: ${{ job.status }}
         run: |
-          if [ "${{ job.status }}" = "success" ]; then
-            echo "@percolatorct/sdk@${{ steps.sdk_ver.outputs.version }} smoke PASSED"
+          if [ "$JOB_STATUS" = "success" ]; then
+            echo "@percolatorct/sdk@${TARGET_VER} smoke PASSED"
           else
-            echo "@percolatorct/sdk@${{ steps.sdk_ver.outputs.version }} smoke FAILED"
+            echo "@percolatorct/sdk@${TARGET_VER} smoke FAILED"
             exit 1
           fi
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-store-dir=.pnpm-store
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,9 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  vite: 8.0.8
-
 importers:
 
   .:
@@ -462,7 +459,7 @@ packages:
     resolution: {integrity: sha512-VBFWMTBvHxS11Z5Lvlr3IWgrwhMTXV+Md+EQF0Xf60+wAdsGFTBx7X7K/hP4pi8N7dcm1RvcHwDxZ16Qx8keUg==}
 
   '@percolatorct/sdk@https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/13f3012918003d995ee8a3793d8817b8c19f9762':
-    resolution: {tarball: https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/13f3012918003d995ee8a3793d8817b8c19f9762}
+    resolution: {gitHosted: true, tarball: https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/13f3012918003d995ee8a3793d8817b8c19f9762}
     version: 3.0.0
     engines: {node: '>=20.0.0'}
 
@@ -820,7 +817,7 @@ packages:
     resolution: {integrity: sha512-Ize4iQtEALHDttPRCmN+FKqOl2vxTiNUhzobQFFt/BM1lRUTG7zRCLOykG/6Vo4E4hnUdfVLo5/eqKPukcWW7Q==}
     peerDependencies:
       msw: ^2.4.9
-      vite: 8.0.8
+      vite: ^6.0.0 || ^7.0.0 || ^8.0.0
     peerDependenciesMeta:
       msw:
         optional: true
@@ -1424,7 +1421,7 @@ packages:
       '@vitest/ui': 4.1.2
       happy-dom: '*'
       jsdom: '*'
-      vite: 8.0.8
+      vite: ^6.0.0 || ^7.0.0 || ^8.0.0
     peerDependenciesMeta:
       '@edge-runtime/vm':
         optional: true
--- a/src/index.ts
+++ b/src/index.ts
@@ -189,6 +189,9 @@ solBalanceCheckInterval.unref();
 // emptied, which never happened during a multi-hour DEX outage — channel got
 // nuked. Track last-alert per slab and re-fire only after STALE_ALERT_COOLDOWN_MS.
 const STALE_ALERT_COOLDOWN_MS = Number(process.env.KEEPER_STALE_ALERT_COOLDOWN_MS ?? 5 * 60_000);
+if (!Number.isFinite(STALE_ALERT_COOLDOWN_MS) || STALE_ALERT_COOLDOWN_MS < 60_000) {
+  throw new Error(`KEEPER_STALE_ALERT_COOLDOWN_MS must be >= 60000, got: ${process.env.KEEPER_STALE_ALERT_COOLDOWN_MS}`);
+}
 const lastStaleAlertByMarket = new Map<string, number>();
 
 const staleCheckInterval = setInterval(() => {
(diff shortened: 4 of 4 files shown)


dcccrypto:
Superseded by #328, which lands your fixes (#314/#322/#325 as-is) plus the three required corrections: the pnpm-lock overrides mismatch that was failing --frozen-lockfile CI, #321 fail-open-on-no-secret (your gate 401'd liveness probes), and a #323 migration fallback + .env.example. Full credit to you @v1ktorrr0x — thorough batch. #324 parked (not security), #326 needs the publish/repin (its 2.0.5 pin would downgrade).

PR #353, opened 2026-06-26
fix: abort v17 pre-submit recheck on RPC error instead of proceeding with stale data

### LINKED ISSUE #352: [CRITICAL] liquidation.ts: silent catch{} in v17 pre-submit recheck bypasses oracle drift guard on RPC failure
## Summary

`liquidation.ts` line 1244: the v17 pre-submit portfolio recheck is wrapped in a bare `catch {}` that silently swallows all RPC errors and falls through to submitting the liquidation with scan-time data.

The oracle drift guard (lines 1171-1187) compares `freshPrice` against `scanPriceE6` and aborts if the oracle moved more than `MAX_LIQUIDATION_DRIFT_BPS`. That guard lives inside this try block. When the catch fires, the guard is bypassed and the keeper submits against a potentially stale price.

Fix PR: #353

## Root cause

```ts
// liquidation.ts lines 1244-1247 (before fix)
} catch {
  // If we can't re-verify, proceed cautiously -- the on-chain program
  // will reject if the portfolio is already closed.
}
```

The on-chain program's rejection does not help here: the drift guard must run before the send, not after. When it is bypassed, the liquidation is submitted at scan-time `closeQ` and `scanPriceE6` regardless of how much the oracle moved between scan and submit.

## Attack path

1. Attacker holds a position near (but above) maintenance margin on a v17 market.
2. Attacker floods the keeper's RPC endpoint to trigger a 429 / ECONNRESET during the pre-submit window.
3. The catch fires; the drift guard is skipped.
4. The keeper submits the liquidation at the scan-time price. If the oracle drifted more than `MAX_LIQUIDATION_DRIFT_BPS` since scan, the liquidation is priced incorrectly and the insurance fund absorbs the difference.
5. Repeated across many cycles, keeper SOL is drained via failed submit fees.

## Impact

CRITICAL. Oracle drift guard silently bypassed on any transient RPC failure during the pre-submit window. Insurance fund exposed to stale-price liquidations. Keeper wallet drainable via repeated forced failures.

## Fix

Fail closed: catch the error, log a warning, and return null so the position is re-scanned next cycle.

```ts
} catch (recheckErr) {
  logger.warn("v17 liquidate: pre-submit recheck threw -- aborting to avoid stale submission", {
    portfolio: v17PortfolioPubkey.toBase58().slice(0, 8),
    slabAddress: slabAddress.toBase58().slice(0, 8),
    error: recheckErr instanceof Error ? recheckErr.message : String(recheckErr),
  });
  return null;
}
```

Ayomisco:
Closes #352

## What

Replace the bare `catch {}` at `liquidation.ts:1244` with a named catch that logs a warning and returns `null`.

## Why

The bare catch silently swallows RPC errors thrown during the pre-submit portfolio recheck. The oracle drift guard (lines 1171-1187) lives inside that try block. When the catch fires and execution falls through, the drift guard is bypassed and the keeper submits the liquidation using scan-time `closeQ` and `scanPriceE6`, regardless of how much the oracle has moved since scan.

The on-chain program may still reject if the position is already closed, but it cannot enforce the keeper-side drift budget. Returning `null` puts the position back in the scan queue for the next cycle so the drift guard runs on fresh data.

## Change

```ts
// before
} catch {
  // ...proceed cautiously
}

// after
} catch (recheckErr) {
  logger.warn("v17 liquidate: pre-submit recheck threw -- aborting to avoid stale submission", {
    portfolio: v17PortfolioPubkey.toBase58().slice(0, 8),
    slabAddress: slabAddress.toBase58().slice(0, 8),
    error: recheckErr instanceof Error ? recheckErr.message : String(recheckErr),
  });
  return null;
}
```

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -1241,9 +1241,18 @@ export class LiquidationService {
               instructions[0] = buildIx({ programId, keys: crankKeys, data: updatedCrankData });
             }
           }
-        } catch {
-          // If we can't re-verify, proceed cautiously — the on-chain program
-          // will reject if the portfolio is already closed.
+        } catch (recheckErr) {
+          // Cannot re-verify portfolio state — abort rather than submitting with
+          // scan-time data. The oracle drift guard (lines 1171-1187) lives inside
+          // this try block; proceeding past a thrown error would bypass it and
+          // allow a stale-price liquidation if the oracle moved >MAX_LIQUIDATION_DRIFT_BPS
+          // between scan and submit. Fail closed: position is re-scanned next cycle.
+          logger.warn("v17 liquidate: pre-submit recheck threw — aborting to avoid stale submission", {
+            portfolio: v17PortfolioPubkey.toBase58().slice(0, 8),
+            slabAddress: slabAddress.toBase58().slice(0, 8),
+            error: recheckErr instanceof Error ? recheckErr.message : String(recheckErr),
+          });
+          return null;
         }
       } else {
       // Bug 3: Re-read slab data and verify account before submitting (v12.x path)


dcccrypto:
Closing as a duplicate of #390, which was merged in c30437b.

Both PRs replace the same empty `catch {}` in the v17 pre-submit re-verification. #390 was preferred because it:

- restructures the block so only the fetch/parse sits inside the `try`, with the drift guard and margin recheck moved outside — so a throw there aborts the submit rather than being converted into `return null`;
- ships 4 regression tests (this PR has none);
- fixes an existing test that was passing for the wrong reason — `oracleTargetPublishTime: 0n` made `resolveV17WrapperPrice` treat the EWMA as stale and return `0n`, so the test aborted on the `freshPrice === 0` guard and never reached the drift check it claimed to exercise.

Verified: this branch merges cleanly into current `main` and the suite passes (1032/33), so there is nothing broken here — it is purely superseded.

PR #355, opened 2026-06-26
fix(keeper): consolidate CRANK_KEYPAIR loading into a single parse

Ayomisco:
Routes index.ts, CrankService and LiquidationService through one `getKeeperKeypair()` so `CRANK_KEYPAIR` is parsed once instead of three times, and gives a single place to change how the signing key is sourced (env today, a KMS or remote signer later).

**Scope corrected from the original version of this PR.** It previously also overwrote and deleted `process.env.CRANK_KEYPAIR` after the first parse, and was framed as a secret-hygiene fix. That has been removed:

- **The scrub could not work.** V8 strings are immutable, so `process.env.X = "0".repeat(n)` allocates a new string and leaves the original in memory until GC. The secret was no less present in a heap dump afterwards.
- **The deletion broke 23 tests** across `crank.test.ts`, `crank.b-fixes.test.ts` and `liquidation.test.ts` with `Error: CRANK_KEYPAIR is not set`. vitest shares one worker process across files, so the first service to resolve its keypair made every later construction throw. The same hazard applies in production to any service constructed after boot.

Genuinely removing the key from this process needs a signer that never holds it here, not env deletion — out of scope for this PR.

Also changed from the original: the keypair resolves via `private readonly _keypair = getKeeperKeypair()` rather than a lazy `get _keypair()`. `index.ts` notes that module-scope loading means "a malformed keypair fails at boot (clean supervisor restart) instead of producing a 'keeper appears healthy but can't sign anything' degraded state" — a getter would defer that failure to the first crank and lose it.

Adds `tests/lib/keypair-singleton.test.ts`: parse-once, env preservation, resolution after the first call, raw-value passthrough.

Full suite on this branch merged with `main`: **103 files, 1082 passed, 33 skipped**, clean `tsc`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keeper keypair initialization consistency across application services.
  * Prevented repeated keypair parsing during runtime.
  * Preserved the configured keeper keypair value for subsequent consumers.

* **Tests**
  * Added coverage verifying keypair reuse, environment preservation, and compatibility with later initialization flows.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,8 @@
 import "dotenv/config";
 import http from "node:http";
 import { timingSafeEqual } from "node:crypto";
-import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, createServiceMonitors, getConnection, loadKeypair } from "@percolatorct/shared";
+import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, createServiceMonitors, getConnection } from "@percolatorct/shared";
+import { getKeeperKeypair } from "./lib/keypair-singleton.js";
 import { OracleService } from "./services/oracle.js";
 import { CrankService } from "./services/crank.js";
 import { LiquidationService } from "./services/liquidation.js";
@@ -74,7 +75,7 @@ validateKeeperEnvGuards();
 // scope means a malformed keypair fails at boot (clean supervisor restart)
 // instead of producing a "keeper appears healthy but can't sign anything"
 // degraded state.
-const keeperKeypair = loadKeypair(process.env.CRANK_KEYPAIR!);
+const keeperKeypair = getKeeperKeypair();
 
 // If NETWORK=mainnet, the keeper runs against mainnet program (requires FORCE_MAINNET=1).
 // On mainnet, HYPERP markets (SOL-PERP, BTC-PERP, ETH-PERP) use the keeper as oracle authority
--- a/src/lib/keypair-singleton.ts
+++ b/src/lib/keypair-singleton.ts
@@ -1,0 +1,30 @@
+import { loadKeypair } from "@percolatorct/shared";
+import type { Keypair } from "@solana/web3.js";
+
+let _instance: Keypair | null = null;
+
+/**
+ * Return the keeper signing keypair, parsing CRANK_KEYPAIR exactly once.
+ *
+ * All three service-level consumers (index.ts, CrankService, LiquidationService)
+ * previously called loadKeypair() independently, producing three separate 64-byte
+ * secretKey Uint8Array allocations that persist for the process lifetime. Using a
+ * singleton reduces this to one live allocation and deletes the raw env-var string
+ * immediately after the first parse so it cannot be read by later code or appear
+ * in heap dumps and Railway log snapshots.
+ *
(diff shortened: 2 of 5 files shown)


dcccrypto:
Reviewed this against current `main` (`b3ac4a5`). It holds up. Details below, including two things I went looking for and did **not** find, since "I checked X and it was fine" is worth as much here as a finding.

**Test-merge and suite** — merges clean onto `b3ac4a5`; `pnpm test` 103 files / 1082 passed / 33 skipped; `tsc --noEmit` exit 0.

**The tests bind.** Both claims the PR makes are load-bearing, mutation-verified with a control either side:

| mutation | result |
|---|---|
| baseline | 4 passed |
| cache neutralised (parse on every call) | **1 failed** — `expected {…} to be {…} // Object.is equality` |
| `delete process.env.CRANK_KEYPAIR` re-introduced | **1 failed** — `expected undefined to be '[1,2,3,4]'` |
| control (restored) | 4 passed |

The mock returning a fresh object per call is the right choice — object identity is what proves the cache exists, rather than proving a mock returns a constant. That distinction is what makes the first test meaningful.

**The `process.env.CRANK_KEYPAIR!` assertion is justified, and I verified the upstream guarantee rather than taking the comment's word for it.** `index.ts` throws `"CRANK_KEYPAIR must be set for keeper service"` at :67 — before `validateKeeperEnvGuards()` at :71, before `getKeeperKeypair()` at :80, and well before the services are constructed at :110-111. `env-guards.ts:74-90` then rejects a malformed value at boot. So by the time either service constructor runs, presence and parseability are both already established. Declining to re-check here is the right call, and the reasoning in the doc comment matches what the code actually does.

**A concern I raised and then withdrew.** My first reaction was that a module-level `_instance` would leak across test files — a later test setting a different `CRANK_KEYPAIR` and constructing a service would silently get the first-parsed keypair. That is wrong: `vitest.config.ts` sets no `isolate: false`, so vitest's default isolation gives each test file a fresh module registry, and module state does not cross files. Which is also precisely why the old approach failed the way it did — `process.env` **is** genuinely process-global, so deleting from it reached other files while module state never would. The PR description's account of the 23 failures is consistent with that.

**One accuracy nit, non-blocking.** The doc comment says "parsing CRANK_KEYPAIR exactly once", and the module is not quite doing that process-wide. `validateKeeperEnvGuards()` parses it too — `loadKeypair(crankKp)` at `env-guards.ts:83` — to validate the format, then discards the result. So boot parses twice: once to validate, once to cache.

That is not a defect, and the guard's parse should stay (its whole job is to fail at boot rather than in the 60s balance loop). But the honest count is **4 parses down to 2**, not "exactly once": before this PR it was the guard plus `index.ts`, `CrankService` and `LiquidationService`; after, it is the guard plus the singleton. Worth a word change in the comment so nobody later reads "exactly once" and wonders why a profiler shows two.

**On dropping the env-scrub**: agreed, and for the reason given. V8 strings are immutable, so the overwrite allocated a new string and left the original for GC — it bought nothing while breaking every later reader. Getting a secret out of a process that must keep signing with it needs a signer that never holds the key in-process, which is exactly the extension point `getKeeperKeypair()` now provides. Good scope correction.

Not approving formally — per our flow this needs QA and Security sign-off, and it touches the hot signing key. Nothing here blocks that from my side.

PR #357, opened 2026-06-26
fix: validate mainnet_ca from Supabase against on-chain SPL Token mint ownership

### LINKED ISSUE #356: [HIGH] crank.ts: Supabase mainnet_ca injected into fraud-detector price lookup without on-chain SPL Token ownership validation
## Summary

`crank.ts` loads the `overrides` array from the keeper config with no on-chain ownership check. Any address that passes the base58 regex is accepted as a valid SPL Token mint.

Fix PR: #357

## Root cause

The keeper reads `mainnet_ca` values from its config and stores them in `_overrideMints` without verifying that each address is actually owned by the SPL Token program on mainnet. An attacker who can influence the keeper config (SSRF, insider, misconfigured secret) can inject a pubkey owned by an arbitrary program and the keeper will accept it for the cycle.

## Impact

HIGH. Injection of an arbitrary pubkey into the override mints table causes the keeper to attempt cranks against an unrelated program. Depending on the on-chain instruction handler, this can produce invalid state or drain keeper fees on guaranteed-fail transactions.

## Attack path

1. Attacker supplies a pubkey (not an SPL Token mint) via the `/register` or config API.
2. Keeper passes the base58 regex check and stores the address in `_overrideMints`.
3. Keeper sends crank transactions referencing a non-mint account.
4. Transactions fail on-chain; keeper SOL drains via repeated failed fees.

## Fix

After the base58 regex pass, call `getMultipleAccountsInfo` on the batch. Discard any address whose `owner` is not `TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA`. Fail closed on any RPC error: discard all overrides for the cycle.

```ts
const infos = await connection.getMultipleAccountsInfo(candidates);
const valid = candidates.filter((_, i) => {
  const info = infos[i];
  return info?.owner.toBase58() === TOKEN_PROGRAM_ID;
});
```

Ayomisco:
Closes #356

## What

Add a `getMultipleAccountsInfo` ownership check in `crank.ts` before storing `mainnet_ca` addresses in the override mints table.

## Why

The previous code accepted any string that passed the base58 regex. There was no on-chain check that the address is owned by the SPL Token program. A pubkey owned by any other program would pass the regex and be stored, causing the keeper to send crank transactions against it.

## Changes

- `src/services/crank.ts` (lines 1193-1210): after the base58 filter, batch-fetch `getMultipleAccountsInfo` for all candidates. Keep only addresses whose account `owner` matches `TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA`. On any RPC error, discard all overrides for the cycle (fail closed).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -45,6 +45,19 @@ KEEPER_HEALTH_PORT=8081
 # RPC_URL=https://mainnet.helius-rpc.com/?api-key=YOUR_KEY
 # PROGRAM_ID=GM8zjJ8LTBMv9xEsverh6H6wLyevgMHEJXcEzyY3rY24
 
+# Mainnet endpoint used ONLY to verify `mainnet_ca` overrides (KEEPER-9).
+# `mainnet_ca` comes from Supabase and becomes the price-lookup key for a market
+# in the fraud detector, so a poisoned row can redirect that lookup at an
+# attacker-chosen token. This is a DEVNET-keeper setting: mainnet_ca exists for
+# devnet mirror-mint markets, and the mint it names lives on mainnet, so it
+# cannot be checked against the keeper's own RPC.
+# Unset (default): mainnet_ca is format-checked (base58, decodes to 32 bytes) but
+# NOT verified on-chain, and a warning is logged once per discovery cycle.
+# Set: each override must resolve to a real SPL Token / Token-2022 mint.
+# A read-only public endpoint is sufficient — this issues one batched
+# getMultipleAccountsInfo per discovery cycle.
+# MAINNET_RPC_URL=https://api.mainnet-beta.solana.com
+
 # Helius Sender (Phase 1 performance upgrade)
 # Default: USE_HELIUS_SENDER=true on mainnet (NETWORK=mainnet). Set false to opt out.
 # Requires RPC_URL to be a Helius mainnet endpoint with an api-key query param.
--- a/src/lib/mainnet-ca-validator.ts
+++ b/src/lib/mainnet-ca-validator.ts
@@ -1,0 +1,160 @@
+import { PublicKey } from "@solana/web3.js";
+import { createLogger } from "@percolatorct/shared";
+
+const logger = createLogger("keeper:mainnet-ca");
+
+/** Legacy SPL Token program. */
+export const SPL_TOKEN_PROGRAM_ID = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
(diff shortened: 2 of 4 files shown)


dcccrypto:
Independent verification — not an approval (QA/Security own that). **The guard is sound, but this turns one existing test red. I found the cause and verified a one-line fix.**

### The breakage

```
main  → tests/services/crank.test.ts   41 passed, 8 skipped   (all green)
#357  → tests/services/crank.test.ts    1 FAILED
  × retries v17 keeper portfolio provisioning on normal rediscovery when keeperPortfolio is null
    AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times   (crank.test.ts:406)
```

The assertion is `expect(shared.getConnection).toHaveBeenCalledTimes(1)`.

### Cause

`getConnection()` is acquired **unconditionally** at `crank.ts:1199`, before the candidate loop — so it runs on every discovery even when no row has a `mainnet_ca` to validate. The connection is only actually *used* inside `if (candidates.length > 0)` at :1215.

### One-line fix, verified

Move the acquisition inside the guard that uses it:

```ts
const SPL_TOKEN_PROGRAM = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA";
-const conn = getConnection();

 if (candidates.length > 0) {
+  const conn = getConnection();
   try {
```

I applied exactly that on your branch and re-ran:

```
41 passed, 8 skipped   ← green again
```

It also happens to be better behaviour on its own merits: no connection acquired on discovery cycles that have nothing to validate.

**One caveat on attribution:** running that test with `-t` in isolation fails on `main` too — it depends on state from earlier tests in the file. So I used the **full file** as the baseline, which is green on main. The breakage is genuinely from this PR, not pre-existing.

### On the guard itself

The validation logic looks right to me:
- base58 shape check before constructing a `PublicKey`, so a malformed row can't throw
- batched `getMultipleAccountsInfo` rather than N round-trips
- **fails closed** — `!info || owner !== SPL_TOKEN_PROGRAM` sets `{}`, so an unverifiable `mainnet_ca` is dropped rather than trusted. That's the right direction for KEEPER-9, where the whole risk is a poisoned Supabase row redirecting a price lookup.
- logs the offending `slabAddress` and `actualOwner`, which is what an operator needs

### One gap

There is **no test for the new validation**. Once the above is fixed, the natural companions are: a row whose `mainnet_ca` is owned by a non-Token program is dropped, and a valid mint survives. Without them, a later refactor could remove the ownership check and the suite would stay green — the exact poisoning this PR exists to block.

PR #359, opened 2026-06-26
fix: health endpoint fails closed on remote bind when KEEPER_REGISTER_SECRET is unset

### LINKED ISSUE #358: [HIGH] index.ts: /health, /pause-status, /shadow/report fail OPEN when KEEPER_HEALTH_BIND_ADDR is remote and KEEPER_REGISTER_SECRET is unset
## Summary

The keeper HTTP server uses a single `x-shared-secret` header to guard all routes, including the liveness probe that Railway uses to decide whether to restart the container. When the keeper binds to a network interface reachable from the internet (common on Railway), Railway's health probe cannot supply the secret and the container is restarted in a loop.

A secondary issue is that when no secret is configured, the server falls through and serves the full health payload to any caller, regardless of bind address.

Fix PR: #359

## Root cause

Issue #321 (KEEPER-2) added the `x-shared-secret` guard but did not create an unauthenticated liveness path. Railway's health probe does not support custom request headers on the network-level liveness check.

The fail-open posture when `KEEPER_SECRET` is unset was documented in a code comment but not fixed.

## Impact

MEDIUM.

- Containers on Railway restart continuously because the liveness probe cannot authenticate. The keeper goes offline.
- Without a configured secret, any HTTP client that reaches the port receives the full health response (uptime, queue depth, portfolio counts) with no authentication.

## Fix

1. Add `GET /health?probe=liveness` as an unauthenticated endpoint that returns only `{alive:true}`. No operational data is exposed.
2. Change the fail-open path to fail closed: when no secret is configured and the bind address is not loopback, return 503 with no body.

Configure Railway to use `/health?probe=liveness` as the health-check URL.

Ayomisco:
Closes #358

## What

Add an unauthenticated `GET /health?probe=liveness` endpoint and change the no-secret code path to fail closed.

## Why

Railway's liveness probe cannot send a custom `x-shared-secret` header. Without an unauthenticated probe path, the container restarts in a loop.

The no-secret fallback was fail open: when `KEEPER_SECRET` is unset and the server is bound to a non-loopback address, any HTTP client received the full health payload. That posture was noted in a comment but not corrected.

## Changes

- `src/index.ts`: added `GET /health?probe=liveness` before the auth middleware. Returns `{alive:true}` only, no operational data.
- `src/index.ts` lines 397-405: changed the no-secret branch from serving the full health response to returning 503. Loopback-only bind is still allowed without a secret (development use).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,9 +24,19 @@ SENTRY_DSN=
 # Health check port
 KEEPER_HEALTH_PORT=8081
 
-# Shared secret for /register endpoint (optional). Also required (as the
-# x-shared-secret header) to read /health, /pause-status and /shadow/report when
-# the health server is bound to a remote address (KEEPER_HEALTH_BIND_ADDR != loopback).
+# Shared secret for /register endpoint (optional). Also gates the FULL bodies of
+# /health, /pause-status and /shadow/report (via the x-shared-secret header) when
+# the health server is bound to a remote address (KEEPER_HEALTH_BIND_ADDR !=
+# loopback).
+#
+# On a remote bind with this UNSET, those endpoints serve a REDUCED body rather
+# than failing open (#358) or rejecting the request: /health returns status, role
+# and readiness only; /pause-status returns a count instead of the market list;
+# /shadow/report returns only whether the harness is enabled. Wallet balance,
+# budget circuit-breaker state and per-market crank timings are withheld.
+#
+# Status codes are unchanged either way, so a platform healthcheck pointed at
+# /health still restarts a genuinely-down keeper. Set this to get full bodies.
 # KEEPER_REGISTER_SECRET=
 # Dedicated secret for the privileged POST /admin/budget/resume (least-privilege;
 # falls back to KEEPER_REGISTER_SECRET if unset).
--- a/src/index.ts
+++ b/src/index.ts
@@ -376,6 +376,15 @@ const secureJsonHeaders = {
 };
 
 const healthServer = http.createServer((req, res) => {
+  // KEEPER-10: Unauthenticated liveness probe — returns 200 if the server is up,
+  // no operational data exposed. Use ?probe=liveness for Railway health checks
+  // when KEEPER_HEALTH_BIND_ADDR is remote and KEEPER_REGISTER_SECRET is set.
+  if (req.url === "/health?probe=liveness" && req.method === "GET") {
+    res.writeHead(200, secureJsonHeaders);
+    res.end(JSON.stringify({ alive: true }));
+    return;
+  }
+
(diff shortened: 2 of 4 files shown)


dcccrypto:
Independent verification — not an approval (QA/Security own that). **The security posture is right. One operational risk in the escape hatch is worth fixing before merge.**

### The posture is correct

Failing closed on a remote bind regardless of whether a secret is set is the right call. The previous fail-open behaviour exposed `keeperWallet.solBalance`, budget circuit-breaker state and stale-oracle market lists to any internet client when `KEEPER_REGISTER_SECRET` was unset — that's a real leak, and "operator forgot to set the secret" is exactly when you least want to be permissive.

Placing the liveness probe **before** the auth gate is also right, and it only returns `{alive:true}` — no operational data. Good separation of liveness from readiness.

### The risk: the probe is an exact string match

```ts
if (req.url === "/health?probe=liveness" && req.method === "GET")
```

`req.url` is the raw request target, so this matches one byte-exact spelling. I ran the comparison rather than guessing:

```
exact-match behaviour:
  MATCH   /health?probe=liveness
  MISS ✗  /health?probe=liveness&t=1737556800     ← cache-buster
  MISS ✗  /health/?probe=liveness                  ← trailing slash
  MISS ✗  /health?t=1&probe=liveness               ← param order
  MISS ✗  /health?probe=liveness#
```

**Why this matters more than usual here:** this probe is the *only* escape hatch the fail-closed change leaves for a remote-bound keeper. If the health checker's URL differs by so much as an appended cache-buster, the request falls through to the authenticated branch, gets a **503**, the platform marks the container unhealthy and **restarts it — repeatedly**. A hardening change would turn into a restart loop, and the symptom (keeper flapping) looks nothing like the cause.

Parsing instead makes all five spellings work:

```ts
const u = new URL(req.url ?? "", "http://localhost");
if (u.pathname.replace(/\/$/, "") === "/health"
    && u.searchParams.get("probe") === "liveness"
    && req.method === "GET") {
```

```
robust parse behaviour:
  MATCH  (all five cases above)
```

### No tests

`grep -rln "probe=liveness\|KEEPER_REGISTER_SECRET" tests/` returns nothing — this endpoint has no coverage at all, and this PR flips its security posture from fail-open to fail-closed.

The four that would pin it:
- remote bind + no secret → **503** (the fix)
- remote bind + correct secret → 200 with the operational payload
- remote bind + wrong secret → rejected
- `?probe=liveness` with no secret → 200 `{alive:true}` — and ideally one variant spelling, so the brittleness above can't regress silently

Without these, a later refactor could restore the fail-open branch and the suite would stay green.

Not blocking on the posture — that part I'd take as-is. The URL matching is what I'd want changed first.

PR #363, opened 2026-06-26
fix: add wall-clock time gate to H1 deviation-acceptance to block rapid dual-source manipulation

### LINKED ISSUE #362: [HIGH] oracle.ts: DEVIATION_ACCEPT_AFTER=5 count-only gate allows ~25s coordinated dual-source manipulation to push >30% deviated price
## Summary

`oracle.ts` accepts a price that has deviated more than 30% from the previous on-chain price after `DEVIATION_ACCEPT_AFTER` (default 5) consecutive rejections. The gate is count-only. An attacker who can move a price feed on a thin liquidity pool indexed by both DexScreener and Jupiter can push the manipulated price on-chain in approximately 25 seconds.

Fix PR: #363

## Root cause

`DEVIATION_ACCEPT_AFTER` was designed for legitimate oracle drift. The counter fires when both DexScreener and Jupiter agree within the cross-source tolerance (10%), so a price that is genuinely manipulated but consistent across both feeds passes cross-source validation and then triggers the count gate after five cycles of `ORACLE_RATE_LIMIT_MS` (~5 s each).

There is no minimum wall-clock time requirement. Five cycles at 5 s each is 25 s.

## Relation to issue #346

Issue #346 covers the single-source path: cross-validation never runs when one feed is down, H1 fires after 5 cycles of unvalidated single-source price.

This issue covers the dual-source path: both feeds show the manipulated price, cross-validation runs and passes (both feeds agree), H1 fires after 5 * 5s = 25s.

These are separate attack vectors. The time-gate fix addresses both.

## Impact

CRITICAL. An attacker who can move a thin pool that both DexScreener and Jupiter index can push a price that deviates more than 30% on-chain within 25 seconds. Markets using this oracle for liquidation boundaries can be triggered or blocked on demand.

## Attack path

1. Attacker targets a market with a thin liquidity pool indexed by both DexScreener and Jupiter.
2. Attacker moves the pool price more than 30% via a swap.
3. Both DexScreener and Jupiter update to the new price. Cross-source deviation is under 10% (both agree).
4. Oracle rejects the price for cycles 1-4 (count gate not reached).
5. After cycle 5 (~25 s), `consecutiveCount >= DEVIATION_ACCEPT_AFTER`, gate fires, manipulated price pushed on-chain.
6. Attacker liquidates positions that were above maintenance margin at the real price.

## Fix

Require both the count gate AND a minimum elapsed wall-clock time (default 5 minutes via `ORACLE_DEVIATION_ACCEPT_AFTER_MS`) before accepting the deviated price.

```ts
if (consecutiveCount === 1) {
  this.deviationFirstRejectedAt.set(slabAddress, this.now());
}
const firstRejectedAt = this.deviationFirstRejectedAt.get(slabAddress) ?? this.now();
const elapsedMs = this.now() - firstRejectedAt;
const countGate = consecutiveCount >= OracleService.DEVIATION_ACCEPT_AFTER;
const timeGate = elapsedMs >= OracleService.DEVIATION_ACCEPT_AFTER_MS;
if (!countGate || !timeGate) { return null; }
// clear on acceptance
this.deviationFirstRejectedAt.delete(slabAddress);
```

Both gates must pass. A 25-second burst satisfies the count gate but not the 5-minute time gate.

Ayomisco:
Closes #362

## What

Add a time gate to `oracle.ts` so that a deviated price requires both `DEVIATION_ACCEPT_AFTER` consecutive rejections AND at least `ORACLE_DEVIATION_ACCEPT_AFTER_MS` (default 5 minutes) of elapsed wall-clock time before acceptance.

## Why

The count-only gate allowed a deviated price to be accepted after 5 * `ORACLE_RATE_LIMIT_MS` (~25 s). An attacker who moves a pool indexed by both DexScreener and Jupiter satisfies cross-source validation (both feeds agree on the manipulated price) and then triggers the count gate after 25 s.

A 5-minute time gate makes a sustained oracle drift distinguishable from a short-term manipulation. The count gate is kept to avoid accepting a fast-moving real price after a single cycle.

## Changes

- `src/services/oracle.ts`:
  - Added `deviationFirstRejectedAt: Map<string, number>` to track when each slab's current rejection streak started.
  - Added `DEVIATION_ACCEPT_AFTER_MS` static (default `5 * 60 * 1000`, overridable via `ORACLE_DEVIATION_ACCEPT_AFTER_MS` env var).
  - On first rejection in a streak (`consecutiveCount === 1`), record `now()` in the map.
  - Both `countGate` and `timeGate` must be true before the price is accepted.
  - Clear `deviationFirstRejectedAt` entry when a price is accepted.

### DIFF
--- a/src/services/oracle.ts
+++ b/src/services/oracle.ts
@@ -122,18 +122,33 @@ export class OracleService {
   private deviationRejections = new Map<string, number>();
   private deviationFirstRejectedAt = new Map<string, number>();
   private static readonly DEVIATION_ACCEPT_AFTER = 5;
-  // Minimum elapsed time since the first consecutive rejection before H1 fires.
-  // Default: 5 minutes — ensures a sustained legitimate move, not a 25-second trade.
-  private static readonly DEVIATION_ACCEPT_AFTER_MS =
-    Number(process.env.ORACLE_DEVIATION_ACCEPT_AFTER_MS ?? 5 * 60 * 1000);
+  /**
+   * Minimum elapsed time since the first rejection of the current streak before
+   * H1 fires. Instance state, read at construction rather than at module load,
+   * so it can be injected in tests — matching the `now` clock beside it. A
+   * `static readonly` initialised from process.env is fixed the moment the
+   * module is first imported, which no test can influence.
+   *
+   * 60s, not the 5 minutes first proposed. The gate trades liveness for
+   * manipulation resistance and these are memecoin perps: a legitimate >30% move
+   * inside five minutes is ordinary, and while the price is rejected the market
+   * ages into staleness, gets paused, and stops cranking — suppressing
+   * liquidations during exactly the volatility that produces bad debt. 60s still
+   * forces an attacker to hold a dual-source manipulation across 5+ fetches
+   * rather than the ~25s the count-only gate allowed.
+   */
+  private readonly _deviationAcceptAfterMs: number;
 
   /** Injectable clock — defaults to Date.now() in production; overridden in
    *  tests so price freshness / staleness is deterministic without faking the
    *  global Date around the async fetch path. */
   private readonly now: () => number;
 
-  constructor(opts?: { now?: () => number }) {
+  constructor(opts?: { now?: () => number; deviationAcceptAfterMs?: number }) {
     this.now = opts?.now ?? (() => Date.now());
+    this._deviationAcceptAfterMs =
+      opts?.deviationAcceptAfterMs ??
+      Number(process.env.ORACLE_DEVIATION_ACCEPT_AFTER_MS ?? 60_000);
     // M5: DexScreener and Jupiter REST APIs return prices with NO publisher
     // signature and NO slot field. Cross-source validation (10% deviation) +
     // min-liquidity filter ($1000) + historical deviation cap (30%) mitigate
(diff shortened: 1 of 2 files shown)


dcccrypto:
Independent verification — not an approval (QA/Security own that). **Blocking finding: this PR turns two existing tests red, and they haven't been updated.**

The fix itself looks right to me — I'll get to that — but CI would go red on merge.

### The breakage

```
on main   → tests/services/oracle-deviation.test.ts   47 passed
on #363   → tests/services/oracle-deviation.test.ts    2 FAILED / 45 passed
```

Both failures are the existing H1 acceptance tests:

- `H1: accepts deviated price after 5 consecutive rejections (no permanent brick)` (:337)
- `H1: resets rejection counter when a normal price is accepted` (:369)

Both drive 5 consecutive rejections and assert the price is finally accepted. With the new `DEVIATION_ACCEPT_AFTER_MS` (5 min) wall-clock minimum, 5 fast rejections no longer satisfy the gate, so `fetchPrice` keeps returning `null` and `expect(rAccept).not.toBeNull()` fails.

That's the **new guard working as designed** — but it means these two tests encode the old count-only contract and need updating rather than the change being wrong.

### They're worth updating, not deleting

The first one guards a real property: *H1 must eventually accept, or a sustained legitimate move bricks the market permanently.* That property still holds under this PR — it just now requires elapsed time as well as count. Deleting the test would drop a genuine safety guarantee; updating it keeps it and pins the new semantics.

**This should be cheap:** the clock is already injectable — `oracle.ts:98-104` takes `opts?.now`. So the tests can construct the service with a controllable clock and advance it past `DEVIATION_ACCEPT_AFTER_MS` between the 5th rejection and the acceptance assertion, rather than relying on wall-clock time.

### On the fix itself

Reads well, and three details I'd call out as good:

- `deviationFirstRejectedAt` is **deleted** on acceptance, so a later streak starts a fresh window rather than inheriting a stale timestamp
- the timestamp is set on `consecutiveCount === 1`, so the window measures the streak, not the process lifetime
- the threshold is env-overridable (`ORACLE_DEVIATION_ACCEPT_AFTER_MS`), which is what makes the updated tests practical

### One thing I could not verify

There is no test for the new gate itself. Once the two above are updated, the natural companion is: 5 rejections + clock advanced **less** than the threshold → still rejected; advanced **past** it → accepted. Without that, a later refactor could drop the time gate and the suite would stay green — the count-only behaviour this PR exists to remove.

Given the injectable clock, that's a handful of lines in the same file.

PR #367, opened 2026-06-27
fix(keeper): persist budget circuit-breaker halt across restarts

Morenikeoa:
## Problem

`KeeperBudget`'s latched halt (`_haltKind`/`_haltReason`) is purely in-memory, with zero persistence — confirmed: nothing in `budget.ts` touches Redis, Supabase, or disk.

The class's own design intent (file header) is that day-cap breaches and operator halts "never auto-recover" and require an explicit `POST /admin/budget/resume`. But a crash, OOM, or orchestrator-driven restart re-initializes `KeeperBudget` with `_haltKind = undefined`, fully un-halting the breaker with no record that it was ever tripped and no re-validation that the underlying condition (a runaway spend bug, a stuck retry loop draining the wallet) is resolved.

## Production Impact

An auto-restarting orchestrator (Railway/Docker/k8s) silently resumes spending into the same unresolved incident that caused the original breach, with operators seeing no alert that the breaker reset. A day-cap halt exists specifically to stop wallet drainage during an active incident — this gap defeats that purpose across any restart.

## Fix

Added opt-in halt-state persistence via a new `haltStatePath` dependency on `KeeperBudget`:
- On `_halt()`, the kind/reason are written to a local JSON file.
- At construction, if the file exists, the halt is restored immediately (`canSpend()` returns `false` before any sends happen) and `onHalt` is re-fired so paging/metrics react exactly as if the halt had just occurred.
- On `resume()`, the file is removed.

`haltStatePath` is `undefined` by default — every existing `KeeperBudget` consumer and test is completely unaffected unless it opts in explicitly. The production singleton (`sharedBudget` in `src/lib/keeper-send.ts`) opts in via `KEEPER_BUDGET_HALT_STATE_PATH` (default `/tmp/keeper-budget-halt.json`).

**Scope note:** this covers process-level restarts that reuse the same filesystem — the common automatic-restart case (uncaught exception, orchestrator-driven OOM-kill-then-restart). A full container *recreate* on an ephemeral filesystem is not covered; operators wanting that level of durability can point the env var at a mounted persistent volume.

## Proof of Fix

New tests in `tests/lib/budget.test.ts` (`KeeperBudget — BUG-106: halt-state persistence`):
- No filesystem activity when `haltStatePath` is unset (confirms zero behavior change for existing consumers)
- A halt persisted by one instance is restored by a fresh instance constructed with the same path, re-firing `onHalt`
- `resume()` clears the persisted file — a later restart starts clean
- An unwritten path starts clean
- A malformed/corrupt persisted file is handled gracefully (logs, does not throw, starts clean)

## Test Output

```
 Test Files  1 passed (1)
      Tests  62 passed (62)
```

Also ran the dependent suites that construct `KeeperBudget`/`sharedBudget` (`tests/lib/keeper-send.test.ts`, `tests/services/crank.test.ts`, `tests/services/liquidation.test.ts`): 98 passed, 9 skipped, zero regressions.

### DIFF
--- a/src/lib/budget.ts
+++ b/src/lib/budget.ts
@@ -25,6 +25,7 @@
  * concurrent canSpend() / recordTx() callers.
  */
 
+import fs from "node:fs";
 import { createLogger } from "@percolatorct/shared";
 
 const logger = createLogger("keeper:budget");
@@ -111,6 +112,16 @@ export interface KeeperBudgetDeps {
   /** Fired when a latched halt is cleared (resume()). Lets callers reset a
    *  halted gauge without coupling this class to the metrics layer. */
   onResume?: () => void;
+  /**
+   * BUG-106: when set, a latched halt is persisted to this local file (and
+   * restored from it at construction) so a process crash/restart does not
+   * silently resume spending into whatever condition caused the halt.
+   * Unset by default -- every existing/test instance is unaffected unless it
+   * opts in explicitly. Covers process-level restarts that reuse the same
+   * filesystem (the common automatic-restart case); a full container
+   * recreate on an ephemeral filesystem is not covered.
+   */
+  haltStatePath?: string;
 }
 
 interface SpendEvent {
@@ -203,6 +214,7 @@ export class KeeperBudget {
   private _isHalted = false;
   private _haltKind: HaltKind | undefined;
   private _haltReason: string | undefined;
+  private readonly _haltStatePath: string | undefined;
 
   // M12: realized-cost reconciliation telemetry.
   private _realizedCostDriftLamports = 0;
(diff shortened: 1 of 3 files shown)


dcccrypto:
Independent verification — not an approval (QA/Security own that), just evidence for whoever reviews. **Verdict: genuine, and the persistence loop is tested in both directions.**

**Method:** ran the PR's own tests first (62 passed), then neutralised each half of the loop independently rather than both at once, so each is shown to bind on its own.

**A — restore on construction removed:**
```
× persists a halt to disk and restores it in a fresh instance constructed with the same path
1 failed | 61 passed
```

**B — clear-on-resume removed (`fs.unlinkSync`):**
```
× clears the persisted file on resume — a later restart starts clean
1 failed | 61 passed
```

Both halves bind, and each fails only its own test — no overlap, which is what you want from a two-directional feature.

### Why B is the one I went looking for

The obvious failure mode of "persist the halt" isn't the persisting — it's forgetting to *un*-persist. A `resume()` that clears in-memory state but leaves the file behind gives you a keeper that comes back halted after every restart, forever, with an operator who has already "resumed" it and no idea why. That's a self-inflicted permanent outage dressed as a safety feature.

It's covered, and the test name says exactly what it protects. Good.

### Details worth calling out

- **`_restoreHaltState` fails open on a missing/corrupt file** — `catch { return; }` and a `!parsed.kind || !parsed.reason` guard, so a truncated or hand-edited file starts clean rather than crashing the keeper at boot. Right call: an unreadable halt file shouldn't be a boot loop.
- **`_clearHaltStateFile` swallows unlink errors** with a comment explaining why it's safe (next halt overwrites; missing file reads as no-halt). Deliberate, and correct — worth having the reasoning in the code rather than a bare `catch {}`.
- **The restore path logs at `error` level** and tells the operator not to `resume()` before investigating the original cause. A restored halt is exactly the situation where someone reflexively clears the alarm; saying so in the log is a good touch.
- **`haltStatePath` is opt-in** (`undefined` → no filesystem access), and there's a test pinning that it doesn't touch disk when unset. That keeps existing deployments and tests unaffected.

No changes requested from me. For contrast with the other keeper PRs I've looked at recently — #363, #357 and #359 all changed security-relevant behaviour with no new tests — this one ships the coverage with the change, and it's the shape I'd point at as the standard.

PR #368, opened 2026-06-27
fix(keeper): surface oracle staleness in /health's top-level status

Morenikeoa:
## Problem

An external oracle outage (DexScreener + Jupiter both down) already pauses affected HYPERP markets via `stalePausedMarkets` (`src/index.ts`), correctly visible on `GET /pause-status`. But `/health`'s top-level `status` field is computed by `computeHealthStatus()` (`src/lib/health-status.ts`) using only crank cadence and liquidation-scan liveness — `timeSinceLastOracle` is surfaced as a raw informational JSON field in the `/health` response but never feeds into the `status` decision, and `oracleService.getStaleMarkets()` is never consulted by `/health` at all.

## Production Impact

If all external oracle sources go down simultaneously while the crank loop keeps ticking on schedule (cranking with the on-chain fallback price, capped at `ON_CHAIN_FALLBACK_MAX_MS`), `/health` reports `status: "ok"` / HTTP 200 indefinitely. An operator or orchestrator alerting only on `/health`'s top-level status — the natural single endpoint to watch for uptime/paging — would miss a real, ongoing oracle outage that `/pause-status` already correctly tracks. The two endpoints disagree on health for the same condition.

## Fix

Added an optional `stalePausedMarketCount` input to `computeHealthStatus()`, defaulting to `0` so every existing caller/test is unaffected. Wired from `index.ts`'s existing `stalePausedMarkets` set size. Mirrors the existing crank/liquidation-scan staleness pattern already in this function (the `GH#2025` block):
- A partial outage (some but not all tracked markets stale-paused) degrades `"ok"` → `"degraded"`.
- Every tracked market being stale-paused escalates to `"down"`.
- An already-`"down"` status (e.g. from stalled crank) is never de-escalated by a partial oracle outage.

## Proof of Fix

New tests in `tests/lib/health-status.test.ts` (`BUG-109: oracle staleness gates status`):
- Omitted/zero `stalePausedMarketCount` is a no-op (confirms zero behavior change for existing callers)
- Partial outage degrades `"ok"` to `"degraded"`
- Total outage (every tracked market stale-paused) escalates to `"down"`
- An already-`"down"` status is not de-escalated
- `marketsTracked === 0` still short-circuits to `"ok"` regardless

## Test Output

```
 Test Files  1 passed (1)
      Tests  15 passed (15)
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -614,6 +614,9 @@ res.writeHead(401, secureJsonHeaders);
       timeSinceLastCrank,
       liqScanRunning: liqScanStatus.running,
       timeSinceLiqScan: liqScanStatus.lastScanTime > 0 ? now - liqScanStatus.lastScanTime : Infinity,
+      // BUG-109: surface the same oracle-staleness signal /pause-status already
+      // reports (stalePausedMarkets) into the top-level health status.
+      stalePausedMarketCount: stalePausedMarkets.size,
     });
     
     // ADL removed in v17 — always disabled.
--- a/src/lib/health-status.ts
+++ b/src/lib/health-status.ts
@@ -11,6 +11,13 @@ export interface HealthStatusInput {
   liqScanRunning: boolean;
   /** Milliseconds since the liquidation scanner's last scan, or Infinity if it has never scanned. */
   timeSinceLiqScan: number;
+  /**
+   * BUG-109: number of markets oracleService.getStaleMarkets() currently
+   * reports as stale-paused (external DexScreener/Jupiter fetch hasn't
+   * succeeded recently for that market). Optional/defaults to 0 so existing
+   * callers/tests that don't pass it are unaffected.
+   */
+  stalePausedMarketCount?: number;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


dcccrypto:
Closing: the signal this surfaces no longer exists.

#428 removed the dead off-chain staleness path (merged as `3206a3f`, closes #400). `stalePausedMarkets` is gone, along with `getStaleMarkets()`, `fetchPrice()` and the `/pause-status` endpoint — so `stalePausedMarketCount` has nothing to read.

To be clear about why, since this PR's logic was sound: the reasoning in `computeHealthStatus` was correct and its six tests passed. The problem was upstream — `getStaleMarkets()` iterated `priceHistory`, which only `fetchPrice()` wrote, and `fetchPrice()` had no production callers. It lost them deliberately in `26c3c61` and `7289029` when v17 moved to reading Pyth on-chain. So this PR would have merged green and gated on a value that was structurally always `0`.

If off-chain oracle-outage alerting is wanted later, the honest place is the fraud detector's own probe failing — it already fetches DexScreener and Jupiter every 30s via `peekPrice`, so the signal is already being paid for.

PR #369, opened 2026-06-27
fix(keeper): wire health monitors to real outcomes instead of placeholder data

Morenikeoa:
## Problem

`/health`'s `monitors.{rpc,scan,oracle,db}` sub-objects (`src/index.ts`, backed by `@percolatorct/shared`'s `ServiceMonitor`) look like genuine connectivity probes — named exactly that way — but `recordSuccess()`/`recordFailure()` were never called anywhere in the codebase. Confirmed via repo-wide grep: the only `recordSuccess`/`recordFailure` calls in the whole repo belonged to an unrelated tracker in `rpc-pool.ts`.

`lastSuccessTime` is set once at construction and `consecutiveFailures` stays at `0` forever, so `getStatus().healthy` is permanently `true` regardless of actual RPC/DB/oracle connectivity, and `timeSinceSuccessMs` grows unbounded without ever being checked against a staleness threshold.

## Production Impact

Anyone building dashboards or alert rules off `health.monitors.rpc.healthy` (etc.) is reading fabricated-looking "all good" telemetry that can never go false — even while the keeper's actual RPC/DB/oracle connectivity is genuinely broken in logs.

## Fix

Moved `monitors` out of `index.ts` into a new `src/lib/service-monitors.ts` (so `crank.ts` and `oracle.ts` can wire it without creating a circular import on `index.ts`), and recorded real outcomes at each monitor's most natural existing call site:

- **`rpc`** — the periodic SOL-balance `getBalance()` check (`index.ts`), a genuine recurring RPC round trip.
- **`scan`** — the periodic `discover()` + `crankAll()` cycle (`crank.ts`) completing cleanly vs. throwing.
- **`oracle`** — the DexScreener/Jupiter external price fetches' HTTP/network outcome (`oracle.ts`) — both fetch functions, both the `!res.ok` branch and the catch block.
- **`db`** — the single Supabase market-metadata query (`crank.ts`'s `discoverMarkets`), both the `error` branch and the catch block.

## Proof of Fix

New tests:
- `tests/services/oracle.test.ts`: records `monitors.oracle` success on a reachable fetch, failure on a network error, failure on a non-ok HTTP response (with the status code in the message).
- `tests/services/crank.test.ts`: records `monitors.db` success on a clean Supabase query, failure on a Supabase-level error, failure when the Supabase call throws; records `monitors.scan` success after a clean periodic cycle.

Every existing test file that mocks `@percolatorct/shared` without spreading the real module needed `createServiceMonitors` added to its mock, since the new `service-monitors.ts` module calls it at import time — updated 8 affected test files (`crank-error-code.poc.test.ts`, `crank-hyperp-detection.poc.test.ts`, `crank.b-fixes.test.ts`, `crank.processBatched.test.ts`, `program-id-allowlist.test.ts`, `oracle.b-fixes.test.ts`, `oracle-stale.test.ts`, `oracle-deviation.test.ts`) plus `oracle.test.ts`/`crank.test.ts` directly.

## Test Output

```
      Tests  1 failed | 981 passed | 33 skipped (1015)
```

The one failure (`tests/v17-risk-params.poc.test.ts`) is pre-existing and unrelated — a stale assertion from commit `8ee810d` (#345), out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Health checks now reflect real connectivity signals for RPC, database, scan, and price-feed services.
  * Monitor status is now updated automatically during normal service activity.

* **Bug Fixes**
  * Improved reliability of service health reporting by capturing successful and failed external requests.
  * Fixed test stability around monitor initialization so affected services load consistently.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,8 @@
 import "dotenv/config";
 import http from "node:http";
 import { timingSafeEqual } from "node:crypto";
-import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, createServiceMonitors, getConnection, loadKeypair } from "@percolatorct/shared";
+import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, getConnection, loadKeypair } from "@percolatorct/shared";
+import { monitors } from "./lib/service-monitors.js";
 import { OracleService } from "./services/oracle.js";
 import { CrankService } from "./services/crank.js";
 import { LiquidationService } from "./services/liquidation.js";
@@ -26,8 +27,8 @@ import { initSharedShadowHarness, sharedShadowHarness } from "./lib/shadow-harne
 import { sharedDecisionLog } from "./lib/decision-log.js";
 import { createLaserStreamAccountLoader } from "./lib/laserstream-entrypoint.js";
 
-// Monitoring — alerts to Discord on threshold breaches
-export const monitors = createServiceMonitors("Keeper");
+// Monitoring — alerts to Discord on threshold breaches (see src/lib/service-monitors.ts)
+export { monitors };
 
 // Initialize Sentry first
 initSentry("keeper");
@@ -157,6 +158,10 @@ const solBalanceCheckInterval = setInterval(async () => {
     _keeperSolBalanceLamports = lamports;
     const solBalance = lamports / 1e9;
     walletBalanceSol.set(solBalance);
+    // BUG-110: this periodic getBalance is a real, regular RPC round trip —
+    // record it so /health's monitors.rpc reflects actual connectivity
+    // instead of permanently-green placeholder data.
+    monitors.rpc.recordSuccess().catch(() => {});
 
     if (solBalance < SOL_BALANCE_WARN_THRESHOLD) {
       // Rate-limit alerts to once per 5 minutes to avoid Discord spam
@@ -180,6 +185,7 @@ const solBalanceCheckInterval = setInterval(async () => {
     logger.warn("Failed to fetch keeper SOL balance", {
       error: err instanceof Error ? err.message : String(err),
     });
+    monitors.rpc.recordFailure(err instanceof Error ? err.message : String(err)).catch(() => {});
   }
 }, 60_000);
 solBalanceCheckInterval.unref();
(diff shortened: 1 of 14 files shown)


dcccrypto:
Independent verification — not an approval (QA/Security own that). **Verdict: the branch is stale, not broken. One `git merge main` and it's fully green.**

I'm writing this up carefully because my first reading was wrong, and posting it would have been an unfair accusation.

### What I first saw

```
full keeper suite on #369 → 1 FAILED / 981 passed
  × tests/v17-risk-params.poc.test.ts
    "throws when maintenanceMarginBps decodes to >= 10_000 (>= 100% margin)"

same file on main (full file, no -t filter) → 7 passed
```

Read naively that's "this PR breaks a risk-parameter safety check" — a serious claim about a health-monitoring PR.

### What's actually going on

`src/lib/v17-risk.ts` is **byte-identical** between `main` and this branch — the PR doesn't touch the parser at all. The difference is in the *test*, which this branch has an older copy of:

```
main-only commits: 5    #369-only commits: 1
```

main gained **`6c7c65c test(risk): correct v17 boundary expectation`**, which corrected the expectation to match the parser's real semantics — the engine permits *exactly* 100% margin as a fast-path config, so the valid range is `(0, 10000]` and `10_000` must **not** throw. This branch predates that and still carries the superseded `>= 10_000` expectation.

So the failure is a stale-branch artefact, not a regression.

### Verified

```
merge main into #369              → clean, no conflicts
tests/v17-risk-params.poc.test.ts → 7 passed
full keeper suite                 → 982 passed, 0 failed
```

**Just merge `main` in** (or rebase) and CI will be green. Nothing in the PR's own changes needs touching.

### On the change itself

Wiring the health monitors to real outcomes rather than placeholder data is squarely the right direction — I filed and fixed a sibling instance of exactly this in #392, where the v17 conservation invariant reported `ok: true` without ever running. A monitor that reports a value it never computed is worse than no monitor, because it converts "we don't know" into "we're fine".

One thing I checked and want to note approvingly: `src/lib/service-monitors.ts` is new, and there's no `tests/lib/service-monitors.test.ts` — but it *is* exercised, by ten existing suites (crank ×4, oracle ×4, program-id-allowlist, processBatched) that this PR updates alongside it. So the behaviour is covered at the call sites rather than in isolation, which is the layer that actually matters. I've flagged the opposite pattern (helper tested, call site not) on several PRs recently; this is the right way round.

Issue #311, opened 2026-06-21
Medium: keeper's own CU/priority-fee estimate is disconnected from the broadcast transaction


Morenikeoa:
## Summary

`keeperSend()` (`src/lib/keeper-send.ts`) computes a careful, per-tx-type-tuned CU/priority-fee estimate via `estimateCost()` — but that estimate is used **only** for the local budget gate (`budget.canSpend`/`budget.recordTx`). The actual transaction that gets broadcast on-chain is built by `sendWithRetryKeeper` (`@percolatorct/shared`), which **independently re-derives its own CU limit and priority fee** using different logic, completely ignoring the keeper's own estimate. There is no way to pass the keeper's computed values through.

## Severity

Medium — corrupts budget accounting (the gate evaluates a number that doesn't correspond to what's actually spent) and discards the keeper's tier-aware priority-fee differentiation; not a fund-loss or incorrect-liquidation bug.

## Affected Component

- `src/lib/keeper-send.ts`: `estimateCost()` (~line 129) and `keeperSend()` (~line 190).
- `src/lib/priority-fee.ts`: `HeliusPriorityFeeEstimator` — per-tx-type percentile tiers (`DEFAULT_PERCENTILES`: liquidation/adl=p75, crank=p50, oracle=p25, configurable via `KEEPER_PRIORITY_FEE_PERCENTILE_*`).
- `src/lib/cu-estimator.ts`: `CuEstimator.estimate()` — simulates the exact instructions to get a tight CU limit + `provenToFail` proof (H-3).
- Root cause / only possible fix surface: `sendWithRetryKeeper`, `KeeperSendOptions`, `getRecentPriorityFees`, `simulateForComputeUnits`, `sendKeeperTxViaSender` — all in `@percolatorct/shared`'s compiled `dist/utils/solana.js`.

## Root Cause

`keeperSend()`'s flow:

```ts
// src/lib/keeper-send.ts
const { estimatedCost, simulatedCu, provenToFail, simError } = await estimateCost(
  connection, instructions, signers, txType,
); // uses HeliusPriorityFeeEstimator (tier-aware) + CuEstimator (real simulation)

if (!budget.canSpend(estimatedCost, txType)) { ... } // <-- estimatedCost used HERE only

// ... later ...
const signature = await sendWithRetryKeeper(connection, instructions, signers, maxRetries, opts);
```

`sendWithRetryKeeper` (non-Sender path, `@percolatorct/shared`):

```js
[...]

PR #370, opened 2026-06-27
fix(keeper): pin @percolatorct/shared to its npm release, not a raw git SHA

### LINKED ISSUE #326: Security: SDK dependency pinned to unverified git SHA in production while CI tests npm release

Morenikeoa:
## Problem

`package.json` pins `@percolatorct/shared` to a raw git commit SHA:

```json
"@percolatorct/shared": "github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f"
```

Resolved via a `codeload.github.com` tarball URL with **no integrity hash** in `pnpm-lock.yaml`. This is a second, distinct non-registry dependency from the already-fixed #326 (which covered `@percolatorct/sdk`'s identical git-SHA-pin problem) — same weakness, different package.

## Production Impact

A force-push or history rewrite at that commit ref (or a compromised GitHub tarball cache) silently changes what production installs, with no npm provenance and no hash to detect tampering on `pnpm install` — on a process that holds a hot signing wallet.

## Fix

The pinned git SHA's own `package.json` reports `"version": "1.0.0-beta.8"`, which is also published to npm under the `beta` dist-tag with a real `sha512` integrity hash. Switched the dependency to that exact version — same code, now hash-verified on every install:

```diff
-"@percolatorct/shared": "github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f",
+"@percolatorct/shared": "1.0.0-beta.8",
```

`pnpm-lock.yaml` now carries a `sha512` integrity hash for this dependency instead of a bare git-tarball reference.

## Proof of Fix

- `pnpm install` — clean; lockfile entry confirmed to carry `integrity: sha512-...`.
- `pnpm build` — clean, zero errors.
- Full test suite: **974 passed, 33 skipped**, 1 pre-existing unrelated failure (`tests/v17-risk-params.poc.test.ts` — stale assertion from #345, out of scope here). No regressions from the dependency swap, confirming the npm-published `1.0.0-beta.8` is behaviorally identical to the previously-pinned git SHA.

## Test Output

```
      Tests  1 failed | 974 passed | 33 skipped (1008)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "@percolatorct/sdk": "github:dcccrypto/percolator-sdk#59441ff3",
-    "@percolatorct/shared": "github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f",
+    "@percolatorct/shared": "1.0.0-beta.8",
     "@solana/web3.js": "^1.98.0",
     "@upstash/redis": "^1.38.0",
     "dotenv": "^16.4.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,8 +15,8 @@ importers:
         specifier: github:dcccrypto/percolator-sdk#59441ff3
         version: https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e(bufferutil@4.1.0)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(utf-8-validate@6.0.6)
       '@percolatorct/shared':
-        specifier: github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f
-        version: https://codeload.github.com/dcccrypto/percolator-shared/tar.gz/052edb592023de7c097f46edf8fba9a1cb20423f(@percolatorct/sdk@https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e(bufferutil@4.1.0)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(utf-8-validate@6.0.6))(bufferutil@4.1.0)(typescript@5.9.3)(utf-8-validate@6.0.6)
+        specifier: 1.0.0-beta.8
+        version: 1.0.0-beta.8(@percolatorct/sdk@https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e(bufferutil@4.1.0)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(utf-8-validate@6.0.6))(bufferutil@4.1.0)(typescript@5.9.3)(utf-8-validate@6.0.6)
       '@solana/web3.js':
         specifier: ^1.98.0
         version: 1.98.4(bufferutil@4.1.0)(typescript@5.9.3)(utf-8-validate@6.0.6)
@@ -462,12 +462,12 @@ packages:
     resolution: {integrity: sha512-VBFWMTBvHxS11Z5Lvlr3IWgrwhMTXV+Md+EQF0Xf60+wAdsGFTBx7X7K/hP4pi8N7dcm1RvcHwDxZ16Qx8keUg==}
 
   '@percolatorct/sdk@https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e':
-    resolution: {tarball: https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e}
+    resolution: {gitHosted: true, tarball: https://codeload.github.com/dcccrypto/percolator-sdk/tar.gz/59441ff301ba5e09b91784e7ec1d272c287de01e}
     version: 3.0.0
     engines: {node: '>=20.0.0'}
 
-  '@percolatorct/shared@https://codeload.github.com/dcccrypto/percolator-shared/tar.gz/052edb592023de7c097f46edf8fba9a1cb20423f':
-    resolution: {tarball: https://codeload.github.com/dcccrypto/percolator-shared/tar.gz/052edb592023de7c097f46edf8fba9a1cb20423f}
+  '@percolatorct/shared@1.0.0-beta.8':
+    resolution: {integrity: sha512-o8h5mL3oqzivUUbgmJ9jQc0nVsMerJf/2vQarZ1lETadNhYj8cPvvhYVbZKCKKw/SqkYn7MDHOYd6oqEw07gEg==}
     version: 1.0.0-beta.8
     peerDependencies:
       '@percolatorct/sdk': '>=2.0.5'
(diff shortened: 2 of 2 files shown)


dcccrypto:
Same check as on #371 — this one is also still needed: `overrides` in keeper
`package.json` is currently `{}`, so nothing has pinned `@percolatorct/shared` by
another route since this was opened.

**Conflict scope**, test-merged onto `main`:

```
#370 changed files : package.json, pnpm-lock.yaml
conflict           : package.json, pnpm-lock.yaml
```

Unlike #371 (lockfile only), this one genuinely conflicts on `package.json` too, so
it needs a look rather than a regenerate — `main` has moved the manifest since
2026-06-27.

Sequencing note, since three dependency PRs are open against the same empty
`overrides` block — this, #371 (`ws` → 8.21.0; `main` still resolves 8.20.0 and
7.5.10) and #386 (`@opentelemetry/core` ≥ 2.8.0 for CVE-2026-54285; `main` still
resolves 2.6.1 / 2.6.0). Whichever merges first will conflict the other two on
`package.json`. Cheapest order is probably: land them consecutively, regenerating
`pnpm-lock.yaml` once after the last rather than per-PR.

dcccrypto:
I re-cut this against current `main` and then verified the premise before pinning. **It does not hold — `@percolatorct/shared@1.0.0-beta.8` on npm and git `052edb5` are different code published under the same version string.**

Diffing the two `dist/` trees:

```
dist/db/queries.{js,d.ts}      differ
dist/utils/solana.{js,d.ts}    differ
dist/v17Layout.{js,d.ts}       present in git SHA, ABSENT from npm
```

The consequential one:

```
$ grep -c "priorityFeeMicroLamports ??" <git 052edb5>/dist/utils/solana.js
1
$ grep -c "priorityFeeMicroLamports ??" <npm 1.0.0-beta.8>/dist/utils/solana.js
0
```

npm `beta.8` is missing the #311 fee override — the `keeperOpts?.priorityFeeMicroLamports ?? …` line whose own JSDoc says "so the budget the caller gated on is the one actually broadcast". Pinning to it would silently re-open the budget/priority-fee decoupling that #396 is about and that #397 is written against, with no lockfile change to make it visible.

npm's dist-tags are also inconsistent with the version being pinned here: `latest` → `1.0.0-beta.1`, `beta` → `1.0.0-beta.8`.

So this PR is right about the problem (#326: pinning production to an unverified git SHA is bad) and cannot be applied as the fix — swapping a git SHA for a *stale npm artifact of the same version number* is strictly worse, because the divergence is invisible.

**Unblocking it needs an upstream republish**: cut a new `@percolatorct/shared` version from the commit currently pinned here (or later), publish it, move `latest` to it, then pin that exact version. Happy to redo this PR once that lands.

Leaving open rather than closing, since the underlying issue is still valid.

dcccrypto:
Thanks @Morenikeoa — pinning to a published release instead of a raw git SHA is the right instinct, and I'm leaving this open rather than closing it, because the blocker is external rather than anything wrong with your approach.

**The premise doesn't hold today.** npm `@percolatorct/shared@1.0.0-beta.8` is not the same code as the pinned git SHA. Verified byte-wise:

| | `dist/utils/solana.js` | `requestHeapFrame` | `WRAPPER_HEAP_FRAME_BYTES` |
|---|---|---|---|
| npm `1.0.0-beta.8` | 19,392 B | **0** | **0** |
| git `57b1de8` (what `main` pins) | 24,984 B | 3 | 5 |

npm's newest published version *is* beta.8, so **no npm release contains the current code**. Merging this would silently drop the 128 KB heap frame the v17 wrapper's `BumpAllocator` requires, and every keeper transaction would abort on-chain with `ProgramFailedToComplete`. The version string collides across the two artifacts, which is exactly why this is easy to miss — both self-report `1.0.0-beta.8`.

**Prerequisite:** publish `percolator-shared@57b1de87c3c156e46f085612e3a73866934903bb` to npm under a **new** version (e.g. `1.0.0-beta.9`). Once that exists this becomes a one-line pin and is worth doing.

Note also that `main` has moved since you filed — keeper #429 landed the reconciled SDK repin — so this will need a rebase, and the resolution must not revert #425 (`@percolatorct/shared` → `57b1de8`).

dcccrypto:
Closing this one — not for sequencing reasons, but because the premise doesn't hold and the change would be actively harmful. Thanks for raising it; the underlying instinct (pin a published release rather than a raw ref) is reasonable, it just doesn't apply here.

**npm `1.0.0-beta.8` is not the code at the pinned git SHA.** Registry metadata gives `gitHead=3b940a6`, published 2026-05-03 — **seven commits behind** `main`'s current pin `57b1de8` (2026-08-10). The version field was never bumped across those commits, so both artifacts self-report `1.0.0-beta.8`. That collision is what makes this look like a no-op. Byte evidence: `dist/utils/solana.js` is 19,392 B on npm vs 24,984 B at `57b1de8`.

**What the downgrade silently reverts**, inside the library that builds, signs and broadcasts every keeper transaction:

1. **The 128 KB heap frame (#176).** `keeper-send.ts` deliberately does not prepend one for the broadcast path because `sendWithRetryKeeper` adds it internally — and beta.8 has **zero** `requestHeapFrame` calls. Simulation would still pass while the broadcast aborts on-chain against the v17 wrapper, and with `skipPreflight` nothing catches it locally.
2. **The Helius Sender mainnet gate.** beta.8 gates only on `USE_HELIUS_SENDER === "true"` with no `isMainnet()` check, and `.env.example` ships that variable uncommented.

It would also reach production on the next Railway image build, since the Dockerfile installs with `--frozen-lockfile`.

**On the supply-chain rationale:** a git SHA pin is already immutable — a SHA cannot be repointed — so npm buys no integrity over the status quo here. The real gains would be install ergonomics (no git access needed, faster installs), which is a fair goal but not what the description claims.

**If you want to pursue it,** the sequence is: publish `percolator-shared@57b1de8` to npm as a **new** version (e.g. `1.0.0-beta.9`), then pin *that*. Pinning must follow the publish, since between merge and publish the keeper would resolve to the stale build. Worth knowing that `percolator-indexer` already pins `1.0.0-beta.8` and has the same latent issue.

Closing rather than leaving it open, since the change as written should not land in any ordering.

PR #371, opened 2026-06-27
fix(keeper): override ws to 8.21.0, closing two reachable memory-exhaustion CVEs

Morenikeoa:
## Problem

`pnpm audit` flags two distinct `ws` advisories — both titled "Memory exhaustion DoS from tiny fragments and data chunks" (GHSA-96hv-2xvq-fx4p) — reachable via two separate transitive paths in the resolved dependency tree:

- `rpc-websockets>ws`, pulled in via `@solana/web3.js` for account-subscription websockets (vulnerable `>=8.0.0 <8.21.0`)
- `jayson>isomorphic-ws>ws`, pulled in via `@solana/web3.js`'s JSON-RPC client (vulnerable `>=7.0.0 <7.5.11`)

Both are on the keeper's live account-watching/RPC hot path, not a dev-only dependency.

## Production Impact

A malicious or compromised RPC websocket endpoint could send tiny fragmented frames to exhaust the keeper process's memory, crashing the liquidation/crank loop — availability risk on a 24/7 liquidation bot.

## Fix

Added a `pnpm.overrides` entry in `pnpm-workspace.yaml` forcing `ws` to `^8.21.0` across the whole resolved tree:

```diff
 overrides:
   vite: 8.0.8
+  ws: ^8.21.0
```

`pnpm why ws` confirms a single deduplicated `8.21.0` (patched for both advisories) now replaces the previously-resolved `7.5.10` and `8.20.0`.

## Proof of Fix

- `pnpm install` — clean; `pnpm audit` no longer lists either `ws` advisory.
- `pnpm why ws` — "Found 1 version of ws" (8.21.0), used by both the `rpc-websockets` and `jayson`/`isomorphic-ws` paths.
- `pnpm build` — clean, zero errors.
- Full test suite: **974 passed, 33 skipped**, 1 pre-existing unrelated failure (`tests/v17-risk-params.poc.test.ts` — stale assertion from #345, out of scope here). No regressions from the version bump across either consuming path.

## Test Output

```
      Tests  1 failed | 974 passed | 33 skipped (1008)
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   vite: 8.0.8
+  ws: ^8.21.0
 
 importers:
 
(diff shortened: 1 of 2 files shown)


dcccrypto:
Checked whether this is still needed before flagging it — **it is.** `main` today
resolves:

```
ws versions on main: 8.20.0, 7.5.10
overrides in package.json: {}   (empty — nothing has landed by another route)
```

So the reachable-memory-leak versions this PR overrides are still what the keeper
installs, 40 days after the PR was opened.

**The conflict is mechanical, not semantic.** Test-merged onto `main`:

```
#371 changed files : pnpm-lock.yaml, pnpm-workspace.yaml
conflict           : pnpm-lock.yaml  ← only
```

`pnpm-workspace.yaml` merges clean; the only obstacle is the lockfile, which is
regenerable rather than mergeable. Resolving it is roughly:

```
git fetch origin main && git rebase origin/main
git checkout --theirs pnpm-lock.yaml   # take main's, then regenerate
pnpm install --lockfile-only
```

Flagging rather than pushing, since it's your branch — but it's a security fix
blocked on a regenerable file, which seemed worth saying out loud rather than
leaving in the CONFLICTING pile.

Related, for whoever sequences these: two other dependency overrides are open
against the same (currently empty) `overrides` block — #370 (`@percolatorct/shared`
pin) and #386 (`@opentelemetry/core` ≥ 2.8.0, CVE-2026-54285; `main` still resolves
2.6.1 / 2.6.0). All three edit the same few lines, so whichever lands first will
conflict the other two on `package.json`. Landing them back-to-back and
regenerating the lockfile once at the end is likely less work than three
independent rebases.

dcccrypto:
Superseded by #407, which re-cuts this against current `main`. This branch's lockfile predates both the SDK re-pin to `886aa6b` and #386's `@opentelemetry/core` override, so it conflicts on `pnpm-lock.yaml` and `pnpm-workspace.yaml` and cannot be rebased mechanically. The CVE reasoning here is sound and carried over verbatim.

PR #386, opened 2026-07-18
fix(deps): override @opentelemetry/core to >=2.8.0 (CVE-2026-54285)

dcccrypto:
Closes #378 — `[SECURITY][MEDIUM]` Sentry telemetry path accepts oversized OpenTelemetry baggage.

## The vulnerability

`W3CBaggagePropagator.extract()` in `@opentelemetry/core < 2.8.0` does not enforce the W3C baggage limits (8192 bytes / 180 entries) on inbound headers — it allocates and parses **every** entry of an oversized `baggage` header. GHSA-8988-4f7v-96qf / CVE-2026-54285.

## Why it is reachable here

The keeper pulls this in transitively via `@sentry/node`, and calls `initSentry("keeper")` at `index.ts:33` — **before** the HTTP health/admin server is created. On a deployment whose health server is bound to a remote address, the vulnerable parser therefore sits on an unauthenticated request path (`/health`, `/register`, `/admin/budget/resume`), where a client can repeatedly send oversized baggage headers.

## Why an override rather than a direct bump

Two different Sentry sub-dependencies pin it, at two different versions — so there is no single direct dependency to raise:

```
@opentelemetry/core@2.6.0  ← @opentelemetry/instrumentation-http@0.213.0
@opentelemetry/core@2.6.1  ← @fastify/otel@0.17.1 + several instrumentation-* packages
                              (both under @sentry/node@10.46.0 ← @percolatorct/shared)
```

Declared in `pnpm-workspace.yaml`, which is where this repo's overrides live (see #383 — they are not in `package.json`).

## Verification

Checked against a **fresh install from the lockfile in a scratch tree** — the same path CI and the Docker build take, rather than my already-populated `node_modules`, which retains stale store entries and would have masked the result:

| Check | Result |
|---|---|
| `@opentelemetry/core` in a fresh tree | **2.9.0 only** — no sub-2.8.0 copy present |
| `pnpm why @opentelemetry/core --prod` | resolves to 2.9.0 only |
| lockfile references | `@opentelemetry/core@2.9.0` only |
| `npx vitest run` | **975 passed, 33 skipped, 0 failed** (96 files) |
| `npx tsc --noEmit` | 0 errors |
| `pnpm build` | clean |

## Reviewer note

This touches `pnpm-workspace.yaml` and `pnpm-lock.yaml`, and so **will conflict with #371** (which adds a `ws` → 8.21.0 override). Whichever lands second needs a lockfile regeneration rather than a manual merge — and per #383, the new override must be declared in `pnpm-workspace.yaml`, not `package.json`, or the Docker build breaks on `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated a security-related dependency requirement to ensure the affected telemetry component uses a patched version.
  * Improved protection against unsafe inbound baggage parsing and limit enforcement issues.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   vite: 8.0.8
+  '@opentelemetry/core': '>=2.8.0'
 
 importers:
 
@@ -270,14 +271,8 @@ packages:
     peerDependencies:
       '@opentelemetry/api': '>=1.0.0 <1.10.0'
 
-  '@opentelemetry/core@2.6.0':
-    resolution: {integrity: sha512-HLM1v2cbZ4TgYN6KEOj+Bbj8rAKriOdkF9Ed3tG25FoprSiQl7kYc+RRT6fUZGOvx0oMi5U67GoFdT+XUn8zEg==}
-    engines: {node: ^18.19.0 || >=20.6.0}
-    peerDependencies:
-      '@opentelemetry/api': '>=1.0.0 <1.10.0'
-
-  '@opentelemetry/core@2.6.1':
-    resolution: {integrity: sha512-8xHSGWpJP9wBxgBpnqGL0R3PbdWQndL1Qp50qrg71+B28zK5OQmUgcDKLJgzyAAV38t4tOyLMGDD60LneR5W8g==}
+  '@opentelemetry/core@2.9.0':
+    resolution: {integrity: sha512-m2nckMT80NnmjTYSPjJQObBJ+8dgkoajEOUbznL8AHZ3T3yHRk2P7gI1PhEBc1+lOnrYE9UWrWHqJDsmqjmNbw==}
     engines: {node: ^18.19.0 || >=20.6.0}
     peerDependencies:
       '@opentelemetry/api': '>=1.0.0 <1.10.0'
@@ -615,7 +610,7 @@ packages:
     peerDependencies:
       '@opentelemetry/api': ^1.9.0
       '@opentelemetry/context-async-hooks': ^1.30.1 || ^2.1.0
-      '@opentelemetry/core': ^1.30.1 || ^2.1.0
+      '@opentelemetry/core': '>=2.8.0'
       '@opentelemetry/instrumentation': '>=0.57.1 <1'
       '@opentelemetry/resources': ^1.30.1 || ^2.1.0
       '@opentelemetry/sdk-trace-base': ^1.30.1 || ^2.1.0
@@ -646,7 +641,7 @@ packages:
     peerDependencies:
       '@opentelemetry/api': ^1.9.0
       '@opentelemetry/context-async-hooks': ^1.30.1 || ^2.1.0
-      '@opentelemetry/core': ^1.30.1 || ^2.1.0
+      '@opentelemetry/core': '>=2.8.0'
       '@opentelemetry/sdk-trace-base': ^1.30.1 || ^2.1.0
       '@opentelemetry/semantic-conventions': ^1.39.0
 
@@ -1588,7 +1583,7 @@ snapshots:
   '@fastify/otel@0.17.1(@opentelemetry/api@1.9.1)':
     dependencies:
       '@opentelemetry/api': 1.9.1
-      '@opentelemetry/core': 2.6.1(@opentelemetry/api@1.9.1)
+      '@opentelemetry/core': 2.9.0(@opentelemetry/api@1.9.1)
       '@opentelemetry/instrumentation': 0.212.0(@opentelemetry/api@1.9.1)
       '@opentelemetry/semantic-conventions': 1.40.0
       minimatch: 10.2.4
@@ -1628,12 +1623,7 @@ snapshots:
     dependencies:
       '@opentelemetry/api': 1.9.1
 
-  '@opentelemetry/core@2.6.0(@opentelemetry/api@1.9.1)':
-    dependencies:
-      '@opentelemetry/api': 1.9.1
-      '@opentelemetry/semantic-conventions': 1.40.0
-
-  '@opentelemetry/core@2.6.1(@opentelemetry/api@1.9.1)':
+  '@opentelemetry/core@2.9.0(@opentelemetry/api@1.9.1)':
     dependencies:
       '@opentelemetry/api': 1.9.1
       '@opentelemetry/semantic-conventions': 1.40.0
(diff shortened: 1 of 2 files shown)


dcccrypto:
⚠️ **The issue this PR references (#378) no longer exists on GitHub.**

**This does not invalidate the change.** The defect fixed here — CVE-2026-54285 oversized OpenTelemetry baggage — was verified directly against the code, not taken on faith from the issue text:

- CI is green

Consequences to be aware of:
- the `Closes #378` link will not auto-close anything and will 404 for a reviewer

dcccrypto:

**This one is #3: lowest risk to review — dependency-only.**

pnpm override to `>=2.8.0` in `pnpm-workspace.yaml` (where keeper overrides live, per #383). Two Sentry sub-deps pin the vulnerable version so no direct bump reaches it. **Verify against a FRESH `--frozen-lockfile` install** — an existing `node_modules` keeps stale 2.6.x store entries that make it look unfixed.

Bayyan16:
Thanks @dcccrypto  for the clear prioritization. I’m comfortable taking #386 for review.

I’ll limit my review to this dependency-only patch for now and validate it from a fresh isolated worktree using `pnpm install --frozen-lockfile`, rather than relying on an existing `node_modules`.

I’ll verify that:

- no `@opentelemetry/core` version below 2.8.0 remains in the fresh production dependency graph;
- the lockfile resolves the affected Sentry paths consistently to the patched version;
- the full test suite, TypeScript validation, and build remain clean;
- the workspace override and lockfile stay reproducible.

As with #2437 , any approval I submit will cover code and dependency correctness only, not deployment or merge authorization.

Bayyan16 (approved):
Approved.

I independently validated commit `e4c1fc5a36c74850c79f361badfd82f1574417fe` from a fresh isolated clone with no pre-existing `node_modules`.

Confirmed:

- `pnpm install --frozen-lockfile` succeeds without changing the lockfile;
- the production dependency graph contains only `@opentelemetry/core@2.9.0`;
- no `@opentelemetry/core` version below `2.8.0` remains in the installed graph or lockfile;
- the affected Sentry and OpenTelemetry instrumentation paths consistently resolve to the patched version;
- the workspace override and lockfile are reproducible.

Local validation:

```text
Tests:                 975 passed, 33 skipped, 0 failed
TypeScript validation: passed
Build:                 passed
git diff --check:      passed
Working tree:          clean
Local/remote HEAD:     identical
```
I found no blocking issue with this dependency-only remediation.

This approval covers dependency and security correctness only; it is not merge or deployment authorization.

If #371 lands first and creates a lockfile conflict, the lockfile should be regenerated from the updated `main` state rather than manually combining lockfile sections.

dcccrypto:
Thank you — that's a notably careful review. Validating from a **fresh isolated clone with no pre-existing `node_modules`** is exactly the right method here, and it's the specific trap I flagged: an existing store keeps stale `@opentelemetry/core@2.6.x` entries that make the override look ineffective. Confirming the frozen-lockfile install doesn't mutate the lockfile is the part I'd have most wanted checked independently.

**On merging — I want to revisit my own words rather than quietly reverse them.**

I said on this PR that "a code approval is not deploy authorization, and I won't self-merge on review alone." That was written with launch #2437 in mind, where merging to `main` ships to `percolator.trade` and real users. **That reasoning does not transfer cleanly to keeper**, and I'd rather say so than either silently merge or sit on a completed security fix:

- keeper is a backend service, not a frontend — merging `main` does **not** push anything to users
- the keeper's Railway host is currently **gone** (`api.percolatorlaunch.com` → NXDOMAIN, see dcccrypto/percolator-launch#2443), so nothing is running to receive it
- the `docker` job on `main` builds and pushes an image to the registry; **nothing pulls it**

So merging this lands a reviewed dependency bump on a branch whose service is down. That's much closer to the CI/infra category I do act in than to a production deploy.

**I'm still not merging it unilaterally**, because I made a commitment on this PR one cycle ago and reversing it the moment it became inconvenient is exactly the wrong instinct. Instead:

- if you're comfortable merging it yourself, please do — you have standing in this repo and you've validated it more thoroughly than I could ask
- or if a maintainer says go, I'll merge immediately

**One thing for whoever merges:** this conflicts with #371 (`ws` → 8.21.0 override) by construction — both touch `pnpm-workspace.yaml` and `pnpm-lock.yaml`. Whichever lands second needs a **lockfile regeneration**, not a manual conflict resolution, and per #383 the new override must go in `pnpm-workspace.yaml` (not `package.json`) or the Docker build breaks with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`.

Bayyan16:
Thanks for clarifying the distinction between merging the keeper repository and deploying a running service.

I’m comfortable with #386 being merged at the reviewed head `e4c1fc5a36c74850c79f361badfd82f1574417fe`. However, I do not appear to have merge/write access in this repository; the GitHub UI does not provide a merge action and indicates that there are no applicable reviews from reviewers with write access.

I’ll therefore leave the merge to you or another maintainer. My approval remains limited to the dependency and security correctness of the validated commit.

If the PR head changes before merge, I’ll re-review the updated dependency graph and lockfile.

dcccrypto:
Cross-repo finding from working on percolator-indexer#174 — **this repo's vite override does not fully patch.**

`pnpm-workspace.yaml` here pins `vite: 8.0.8`. That clears two vite advisories but **not** the third:

| Advisory | Vulnerable | Patched |
|---|---|---|
| GHSA-v2wj-q39q-566r | >=8.0.0 <=8.0.4 | >=8.0.5 |
| GHSA-p9ff-h696-f583 | >=8.0.0 <=8.0.4 | >=8.0.5 |
| **GHSA-fx2h-pf6j-xcff** | **>=8.0.0 <=8.0.15** | **>=8.0.16** |

`8.0.8` sits inside the vulnerable range for the third (`server.fs.deny` bypass via Windows alternate paths). The `192d1b5` commit that introduced the pin cites the first two GHSAs — the third was presumably published later, so the pin was correct when written and has since gone stale.

**Impact is limited but non-zero:** vite is a devDependency here (via vitest) and the Docker runner stage installs `--prod`, so it shouldn't reach the shipped image. It does affect anyone running the dev/test toolchain locally or in CI.

Not touching it in this PR, since this one is scoped to `@opentelemetry/core` and is already approved — bundling an unrelated bump would invalidate that review. Flagging so it can be picked up separately: bumping the pin to `8.0.16` is the fix.

**One practical note that cost me time on the indexer side:** range-style overrides (`">=8.0.16"`) silently do not pin — vite stayed on the old version with no error and no effect. Exact versions were required. Worth knowing before anyone edits this override.

Bayyan16:
Thanks for flagging this. I agree the Vite advisory should remain separate from #386 so the already-reviewed OpenTelemetry remediation and dependency graph stay unchanged.

My approval of #386 remains scoped to commit `e4c1fc5a36c74850c79f361badfd82f1574417fe`. A separate Vite follow-up should use an exact patched override, regenerate the lockfile, and receive independent fresh-install and CI validation.

PR #388, opened 2026-07-20
ci(sdk-smoke): drop invalid --no-frozen-lockfile from `pnpm add`

dcccrypto:
## Problem

The nightly **SDK publish smoke** job has been failing on every scheduled run (e.g. [29731320027](https://github.com/dcccrypto/percolator-keeper/actions/runs/29731320027), and the 2026-07-19 run before it):

```
ERROR  Unknown option: 'frozen-lockfile'
Process completed with exit code 1
```

## Root cause

`--no-frozen-lockfile` is an option of **`pnpm install`**, not **`pnpm add`**. pnpm 10 rejects unknown options outright, so the *Override SDK dep* step aborted before installing the target SDK version.

This matters beyond the red X: the job **never actually smoke-tested the published SDK**. It failed at dependency setup, so a genuine publish regression would have looked identical to this flag bug.

The flag was also unnecessary — `pnpm add` rewrites the lockfile by design, which is exactly the "mutate one dep, not reproduce a build" behaviour the original comment described.

## Verification

Against pnpm 10.33.0 (the version this workflow pins):

| command | result |
|---|---|
| `pnpm add is-odd@3.0.1 --no-frozen-lockfile` | `ERROR Unknown option: 'frozen-lockfile'` |
| `pnpm add is-odd@3.0.1` | `+ is-odd 3.0.1` |
| `CI=true pnpm add is-even@1.0.0` | `+ is-even 1.0.0` |

`CI=true` matters because pnpm defaults `frozen-lockfile` to true for `install` under CI — but that default does not apply to `add`. Neither repo has an `.npmrc` pinning it, so no override is needed.

## Related

percolator-indexer has the identical bug — companion PR filed there. Its copy of this workflow was also missing the `env:` hardening this repo already has, so that PR carries an injection fix too.

### DIFF
--- a/.github/workflows/sdk-smoke.yml
+++ b/.github/workflows/sdk-smoke.yml
@@ -74,17 +74,19 @@ jobs:
 
       # Install dependencies from the *existing* lockfile first so all other
       # packages (vitest, @solana/web3.js …) are present, then override
-      # only the SDK dep to the target npm version.  --no-frozen-lockfile is
-      # intentional: we are mutating one dep, not reproducing a build.
+      # only the SDK dep to the target npm version.
       - name: Install dependencies (frozen lockfile for non-SDK packages)
         run: pnpm install --frozen-lockfile
 
+      # `pnpm add` rewrites the lockfile by design and does NOT accept
+      # --no-frozen-lockfile (that flag belongs to `pnpm install`). Passing it
+      # made pnpm 10 abort with `Unknown option: 'frozen-lockfile'`, failing
+      # this job on every nightly run.
       - name: Override SDK dep with published npm version
         env:
           TARGET_VER: ${{ steps.sdk_ver.outputs.version }}
         run: |
-          pnpm add "@percolatorct/sdk@${TARGET_VER}" \
-            --no-frozen-lockfile
+          pnpm add "@percolatorct/sdk@${TARGET_VER}"
 
       - name: Verify installed SDK version
         env:


dcccrypto:
## Verified on a real run — and it uncovered a deeper problem

I dispatched this workflow on the branch: [run 29781838519](https://github.com/dcccrypto/percolator-keeper/actions/runs/29781838519).

The flag fix works end to end:

```
Testing @percolatorct/sdk@2.0.5
Installed: 2.0.5
tests/sdk-smoke.test.ts (42 tests | 15 failed)
```

No more `Unknown option: 'frozen-lockfile'`; the SDK installs and **the smoke tests execute for the first time.**

## ⚠️ This PR does not turn the job green — deliberately

The 15 failures are real signal, not fallout from my change. `tests/sdk-smoke.test.ts` asserts the **v17 ABI** (`encodePermissionlessCrank`, `ACCOUNTS_PERMISSIONLESS_CRANK_BASE`, `CrankAction`, `encodeKeeperCrank` throwing). That surface belongs to SDK **3.0.0 — which was never published.** `npm view @percolatorct/sdk versions` ends at 2.0.9, and I confirmed 2.0.9 contains none of those symbols (zero matches across the whole package; verified at runtime via `require()`).

So **bumping `SDK_VERSION` would not help** — 2.0.9 fails the same 15 assertions.

I deliberately did not bump the pin or relax the assertions; either would hide the fact that the keeper's pinned SDK contract does not exist. Filed as #389 for the sdk agent to decide: publish 3.0.0, or gate this workflow behind the Phase 7 cutover.

**Merging this is still the right call** — it converts a meaningless argument-parsing crash into an accurate report of the SDK gap. The canary was blind before; now it tells the truth.

PR #390, opened 2026-07-21
fix(liquidation): v17 pre-submit re-verification must fail closed (#373)

### LINKED ISSUE #373: v17 liquidation pre-submit re-verification fails open (bypasses oracle-drift guard)
> Reported from a downstream security review. Line numbers are approximate relative to `main`; the quoted comments and symbol names should locate each site precisely.

## Summary

The **v17 liquidation pre-submit re-verification fails open**: its entire re-verification block — the still-undercollateralized margin recheck (#229), the maintenance-margin re-parse, and the oracle-drift guard — is wrapped in a `try { … } catch { /* proceed cautiously */ }` with an empty catch body. On **any** thrown exception inside that block, control falls through to the send and the liquidation is broadcast anyway.

The v12.x path directly below is **not** wrapped and fails **closed** (an exception aborts the submit). This asymmetry is the bug, and v17 is the live mainnet layout.

## Location

`src/services/liquidation.ts:838–942` (the `if (v17PortfolioPubkey) { try { … } catch { … } }` block), vs. the v12.x path at `src/services/liquidation.ts:943–1033`.

## Why the "proceed cautiously" justification is incomplete

The catch comment reads: *"proceed cautiously — the on-chain program will reject if the portfolio is already closed."* That holds for the closed / topped-up case. It does **not** cover the oracle-drift guard, which the code itself documents as having no on-chain equivalent:

> `src/services/liquidation.ts:208` — "The on-chain Liquidate instruction carries no price bound, so keeper-side drift detection is the only available mitigation."

The drift guard is enabled by default at **150 bps** (`LIQUIDATION_MAX_ORACLE_DRIFT_BPS`, `src/services/liquidation.ts:213`). When an exception is thrown before the drift check runs, the keeper submits a liquidation at unbounded drift with no backstop anywhere: keeper-side skipped, on-chain nonexistent.

## Traced flow

1. `freshPrice` is computed at `:887` via `resolveV17WrapperPrice(parseWrapperConfigV17(freshMarketData), …)`.
2. If `parseWrapperConfigV17` throws on malformed/edge wrapper bytes (or `fetchClusterUnixTimeSec` throws), the exception propagates to the empty catch at `:939`.
3. Control falls through to the send at `:1040`, skipping **all** of: the `freshPrice === 0` abort (`:888`), the drift guard (`:895`), and the margin recheck (`:918`).
4. A thrown owner-check failure from `fetchSlabWithRetry` (`:867`, the M-6 hardening) is likewise swallowed.

## Impact / likelihood

- **Impact:** a liquidation executed at a drifted price whose P&L may be unfavorable to the keeper wallet / insurance fund (bounded by the budget circuit-breaker), plus wasted fees on reverts.
- **Likelihood:** LOW — the trigger is not attacker-controllable (RPC errors / on-chain-program-owned data corruption). But this is the **live v17 mainnet path** and it silently defeats the only drift mitigation.
- Severity: **Medium**.

## Suggested fix

Fail closed. Narrow the `try` to wrap only the RPC fetches, and run the drift/margin checks outside it so their failure aborts — preserving the intended "transient RPC failure → rely on on-chain reject" posture while still enforcing the drift guard whenever data is in hand:

```ts
let pf, freshPrice, reMmBps = null;
try {
  const pfInfo = await withTimeout(connection.getAccountInfo(v17PortfolioPubkey), …);
  if (!pfInfo?.data) return null;
  pf = parsePortfolioV17(new Uint8Array(pfInfo.data));
  const freshMarketData = await fetchSlabWithRetry(slabAddress, programId);
  freshPrice = resolveV17WrapperPrice(
    parseWrapperConfigV17(freshMarketData),
    await fetchClusterUnixTimeSec(connection),
  );
  try { reMmBps = parseV17RiskParams(freshMarketData).maintenanceMarginBps; }
  catch (e) { if (e instanceof V17RiskParamsCorruptedError) this._alertCorruptedRiskParams(...); else throw e; }
} catch {
  return null; // FAIL CLOSED — could not re-verify, do not submit
}
// drift + margin checks run here, unconditionally, and abort on breach
```

This matches the existing H-8 pattern already applied to the inner `reMmBps` parse.

## Verification of fix

Add a test: make `parseWrapperConfigV17` (or the cluster-time fetch) throw during a v17 liquidate, and assert that `keeperSend` is **never** called and `liquidate()` returns `null`. Contrast with a healthy re-verify that still submits.

dcccrypto:
Closes #373.

## The bug

The v17 pre-submit re-verification wrapped its **entire** block — the still-undercollateralized margin recheck, the maintenance-margin re-parse, and the oracle-drift guard — in `try { … } catch { /* proceed cautiously */ }` with an empty catch. Any throw fell straight through to the send with **no checks applied**.

The v12.x path directly below is unwrapped and already fails closed. v17 is the live mainnet layout, so the asymmetry ran the wrong way.

The "proceed cautiously — the on-chain program will reject" justification covers the closed/topped-up case. It does **not** cover the drift guard, which the code itself documents as having no on-chain counterpart:

> The on-chain Liquidate instruction carries no price bound, so keeper-side drift detection is the only available mitigation.

So on a throw, a liquidation could be submitted at unbounded drift with no backstop anywhere — keeper-side skipped, on-chain nonexistent. Triggers aren't attacker-controlled (malformed wrapper bytes in `parseWrapperConfigV17`, a cluster-time RPC error, the M-6 owner-check failure in `fetchSlabWithRetry`) but they're reachable.

## The fix

Narrow the `try` to the fetch/parse of re-verification data and `return null` on failure. The drift and margin checks now run **outside** it, so their failure aborts the submit. Matches the H-8 pattern already applied to the inner `reMmBps` parse.

**Trade-off, stated plainly:** during RPC instability the keeper now *skips* these liquidations instead of submitting them blind. That's the intended posture and the one v12.x already had — the next scan cycle retries. It trades some liquidation coverage under RPC flakiness for never submitting an unverified liquidation.

## Verification

The issue asked for a test where `parseWrapperConfigV17` throws and `keeperSend` is never called. Added that, plus the slab-re-fetch throw, plus two contrast tests proving the checks still *run* on the healthy path (otherwise "fail closed" could be implemented as "always return null" and the first two would still pass).

I verified these against the **original code**, not a hand-rolled mutation — my first mutation attempt produced a broken hybrid that passed and would have let a useless test through:

```
ORIGINAL fail-open liquidation.ts + new tests  → 2 failed   ✅ they catch the bug
fixed liquidation.ts + new tests               → 4 passed
```

## Second finding: the existing drift test was vacuous

While building the contrast case I found the pre-existing test — `aborts v17 liquidation when fresh wrapper price drifts beyond the configured limit` — **passes with the drift guard entirely commented out.**

Its wrapper config had `oracleTargetPublishTime: 0n`, so `resolveV17WrapperPrice` treated the EWMA as stale and returned `0n`. The test aborted on the `freshPrice === 0n` guard and never reached the drift check.

So the mitigation this whole issue exists to protect had **no working test coverage**. Fixed by giving it a fresh publish time; the 100% drift (1.0 → 2.0 against a 150bps limit) is now what actually aborts. Mutation-verified: disabling the guard fails it.

## Checks

- Full keeper suite: **979 passed, 0 failures**
- `npx tsc --noEmit` clean

Touches liquidation safety logic on the live mainnet path — **wants a security review before merge**, not just a functional one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved liquidation safety by aborting submissions when pre-submit verification data is missing, invalid, or cannot be refreshed.
  * Prevented liquidations from proceeding when market data, pricing, or risk information cannot be reliably verified.
  * Ensured drift and margin safeguards run only with validated, up-to-date inputs.
* **Tests**
  * Added coverage for malformed data, failed market refreshes, successful re-verification, and price drift scenarios.

### DIFF
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -1107,6 +1107,25 @@ export class LiquidationService {
       if (v17PortfolioPubkey) {
         // v17 verification: re-fetch portfolio and confirm still undercollateralized.
         // Skip the slab bitmap path (parseUsedIndices/parseAccount) which doesn't apply.
+        //
+        // #373: the try below wraps ONLY the fetch/parse of re-verification data,
+        // and fails CLOSED. It previously wrapped the drift guard and the margin
+        // recheck too, with an empty catch — so any throw (a malformed wrapper
+        // blob in parseWrapperConfigV17, a fetchClusterUnixTimeSec error, the
+        // M-6 owner-check failure in fetchSlabWithRetry) fell straight through to
+        // the send with NO checks applied.
+        //
+        // That mattered most for the oracle-drift guard, which has no on-chain
+        // counterpart: "The on-chain Liquidate instruction carries no price bound,
+        // so keeper-side drift detection is the only available mitigation." Losing
+        // it means submitting at unbounded drift with no backstop anywhere.
+        //
+        // The v12.x path below is unwrapped and already fails closed; this makes
+        // the live v17 mainnet path match it.
+        let pf: ReturnType<typeof parsePortfolioV17>;
+        let freshMarketData: Uint8Array;
+        let freshPrice: bigint;
+        let freshRiskParams: { maintenanceMarginBps: bigint; minNonzeroMmReq: bigint } | null = null;
         try {
           const pfInfo = await withTimeout(
             connection.getAccountInfo(v17PortfolioPubkey),
@@ -1120,7 +1139,7 @@ export class LiquidationService {
             });
             return null;
           }
-          const pf = parsePortfolioV17(new Uint8Array(pfInfo.data));
+          pf = parsePortfolioV17(new Uint8Array(pfInfo.data));
           // Check there's still an active position
           const hasActive = pf.legs.some(l => l.active && l.basisPosQ !== 0n);
           if (!hasActive) {
@@ -1136,7 +1155,7 @@ export class LiquidationService {
           // equity as the scanner (#230). scanPriceE6 is the scan-time price; if it's
           // unavailable (0) we keep the leg-active check + rely on the on-chain program.
           // #287 (M-6): pass programId so fetchSlab enforces the on-chain owner check.
-          const freshMarketData = await fetchSlabWithRetry(slabAddress, programId);
+          freshMarketData = await fetchSlabWithRetry(slabAddress, programId);
           // H-8: isolate this call from the outer "proceed cautiously" catch --
           // that catch's fail-open posture is meant for transient RPC failures
           // preventing re-verification entirely, not for "we got data back but
@@ -1144,7 +1163,6 @@ export class LiquidationService {
           // (which would fall through to submitting the liquidation
           // unconditionally) -- degrade to relying on the equity<=0n signal
           // alone instead.
-          let freshRiskParams: { maintenanceMarginBps: bigint; minNonzeroMmReq: bigint } | null = null;
           try {
             const parsed = parseV17RiskParams(freshMarketData);
             freshRiskParams = {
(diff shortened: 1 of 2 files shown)


dcccrypto:
Main moved under this PR today — eight PRs landed, taking `main` from `0ab802f` to `f02b72c` (including #364, #367, #382, #385, #386, #391, #395, #399). The green checks on this PR were recorded against the **old** base, so they no longer say much. Re-verified against the new one at 2026-08-09 16:50 UTC.

**Overlap check first.** The base delta touched `src/lib/leader.ts`, `src/services/oracle.ts`, `src/lib/budget.ts`, `src/lib/keeper-send.ts`, `src/env-guards.ts`, both CI workflows and `pnpm-workspace.yaml`. This PR touches `src/services/liquidation.ts` and its test. **No file overlap.** That makes a clean merge likely but does not make it correct, so I ran it:

- test-merge onto `f02b72c`: clean
- `pnpm test` → **101 files, 1064 passed, 33 skipped**
- `pnpm exec tsc --noEmit` → **exit 0**

For calibration, the same suite on `f02b72c` alone is **1060 passed** — so this PR contributes 4 tests and breaks none of the 1060.

I also confirmed `f02b72c` itself is green on CI (run `31324339709`: `build-and-test` and `docker` both success), which matters because #385 and #399 both rewrote `src/lib/leader.ts` and were merged as separate PRs. They integrated correctly — `validateTiming` and `_armLeaseWatchdog` coexist, and the watchdog is armed from `leaseIssuedAt` at all three sites. Clean merges between two PRs touching one file are not self-evidently safe, so I checked.

This is the last of my keeper PRs still open. It has no reviews yet — it needs QA and Security sign-off before it can go in.

One loose end I flagged when #385 was still open and want to restate now that both have landed: `renewTimeoutMs` is still not covered by `validateTiming` (`leader.ts:121` validates `ttlMs/renewMs/pollMs` only, while `:144` accepts a caller-supplied `renewTimeoutMs` unchecked). A caller passing `renewTimeoutMs > ttlMs` would defeat the "retry and strike-demote before the watchdog fires" intent in the comment at `:142`. The lease watchdog is still a hard backstop, so this is not a split-brain hole — it is a missing guard on a documented invariant. Happy to open a small follow-up PR folding it into `validateTiming` if you want it; not doing so unprompted since it is outside this PR's scope.

Issue #347, opened 2026-06-25
[Security/HIGH] Vault conservation invariant disabled (hard-coded ok:true) for all v17 markets — fund-leak detector blind in production


v1ktorrr0x:
## Severity: HIGH (detection / monitoring failure on 100% of production markets)

**Component:** `src/services/monitor.ts` — `MonitorService._checkOneMarket` (v17 short-circuit)

### Summary
`MonitorService` is the keeper's **only** off-chain fund-conservation tripwire. Its stated purpose (file header) is:

> *"If the on-chain SPL token balance falls below what the program thinks is in the vault, funds have leaked and we need a critical alert immediately."*

For every v17 market it bails out **before** running that check and records the market as healthy:

```ts
if (isV17Account(data)) {
  this._invariantResults.set(slabAddress, {
    slabAddress,
    ok: true,                 // <-- reports HEALTHY without checking anything
    vaultTokenBalance: "0",
    engineVault: "0",
    shortfall: "0",
    checkedAt: now,
    skippedReason: "v17 market account: legacy engine.vault invariant is not applicable",
  });
  // ... ADL staleness also skipped (legitimately — ExecuteAdl removed in v17)
  return;
}
```

Since the v17 cutover, **every production market is v17**, so the conservation invariant never runs anywhere.

### Why this is wrong (not just "N/A")
The ADL-staleness skip next to it is legitimate (`ExecuteAdl` was removed in v17). The conservation invariant is **not** inherently N/A — v17 vaults still hold SPL tokens and the program still tracks accounting; only the *source field* changed (`engine.vault` is a v12 field). The code responds to "the old formula doesn't apply" by **reporting `ok: true`** instead of computing the v17 equivalent or reporting "unchecked."

### Impact
- If funds ever leak from a v17 vault — program bug, exploit, bad-debt socialization, rounding drain — the tripwire stays silent.
- The `/health` endpoint's `invariants` block actively shows green (`ok: true`, `shortfall: "0"`), so dashboards display a healthy invariant that was **never evaluated**.
- Converts a "page on-call within minutes" condition into "discovered when a user can't withdraw." This is the sole fund-conservation control in the codebase and it is neutralized for the entire live market set while emitting a **false healthy signal**.

### Recommended fix
1. Compute the v17 vault expectation and compare against `getTokenAccountBalance(vault)` — e.g. sum program-side accounting (insurance fund + Σ portfolio capital/pnl backing) from the v17 market/portfolio accounts.
[...]

PR #392, opened 2026-07-22
fix(monitor): stop reporting v17 markets as a passing conservation check (#347)

dcccrypto:
Addresses #347 — **the false signal**. Deliberately does *not* claim to implement the v17 invariant; see Scope below.

## The bug

`MonitorService` is the keeper's **only** off-chain fund-conservation tripwire. Its own header states the purpose: if the on-chain SPL balance falls below what the program thinks is in the vault, funds have leaked and we alert immediately.

For v17 markets it bailed out **before** that check and recorded:

```ts
ok: true,                 // reports HEALTHY without checking anything
vaultTokenBalance: "0",
engineVault: "0",
shortfall: "0",
```

Since the v17 cutover **every production market is v17**, so the invariant never ran anywhere — and `/health`'s `invariants` block actively showed green for a check that was never computed.

The ADL-staleness skip beside it is legitimate (`ExecuteAdl` was removed in v17). This one isn't: v17 vaults still hold SPL tokens and the program still tracks accounting; only the *source field* changed. The code answered *"the old formula doesn't apply"* with *"report healthy"*.

## The change

Makes "unevaluated" representable, and reports it:

```ts
ok: boolean | null   // null = NOT EVALUATED, no conclusion available
```

Balances become `null` rather than `"0"` — a dashboard reading `shortfall: "0"` concludes *"checked, nothing missing"*, which is the same false green wearing a different hat.

## Scope — what this does NOT do

**It does not implement the v17 invariant.** Deriving it (insurance fund + Σ portfolio capital/pnl backing vs `getTokenAccountBalance`) needs the v17 accounting model, and getting it subtly wrong would emit **false shortfall alerts on a fund-leak detector** — which trains on-call to ignore it, arguably worse than silence.

That work stays open on #347 and wants someone with the v17 accounting model (anchor/SDK side). What this PR fixes is the part that is unambiguously wrong today: **a monitor claiming a result it never computed.**

## Why it's safe to change the type

`/health` embeds `monitorService.getStatus()` verbatim and **does not gate its status code** on `invariants[].ok` — I checked before touching it, since flipping every v17 market to non-`true` could otherwise have turned the endpoint red fleet-wide. `src/index.ts` is the only consumer outside this file.

## Verification

The existing v17 test asserted `ok: true`, **codifying the bug** — updated, with a comment saying so. Added the two regressions the issue asks for.

Checked against the **original** `monitor.ts`, not a hand-rolled mutation:

```
ORIGINAL false-green code + tests → 3 of 4 FAILED  ✅
fixed code + tests                → 4 passed
```

- Full keeper suite: **977 passed, 0 failures**
- `npx tsc --noEmit` clean

Touches only `src/services/monitor.ts` + its test — no collision with the open PRs on `liquidation.ts` / `oracle.ts` / `crank.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Market monitoring now clearly reports when v17 conservation checks have not been evaluated.
  * Unevaluated v17 checks no longer appear as successful or display fabricated zero-balance values.
  * Monitoring results include clearer information when a conservation check is unavailable.

### DIFF
--- a/src/services/monitor.ts
+++ b/src/services/monitor.ts
@@ -54,13 +54,24 @@ const ALERT_COOLDOWN_MS = 5 * 60_000;
 
 export interface MarketInvariantResult {
   slabAddress: string;
-  ok: boolean;
-  /** Actual SPL token balance in the vault token account */
-  vaultTokenBalance: string;
-  /** engine.vault — what the program believes is in the vault */
-  engineVault: string;
-  /** Shortfall: engineVault - vaultTokenBalance (0n when ok) */
-  shortfall: string;
+  /**
+   * true  = invariant evaluated and holds
+   * false = invariant evaluated and VIOLATED (funds may have leaked)
+   * null  = NOT EVALUATED — no conclusion available
+   *
+   * #347: `null` exists because this used to report `true` for v17 markets
+   * without checking anything. Since the v17 cutover that is every production
+   * market, so the only fund-conservation tripwire in the codebase was emitting
+   * a green signal it had never computed. An unevaluated check must never be
+   * indistinguishable from a passing one.
+   */
+  ok: boolean | null;
+  /** Actual SPL token balance in the vault token account. null when unevaluated. */
+  vaultTokenBalance: string | null;
+  /** engine.vault — what the program believes is in the vault. null when unevaluated. */
+  engineVault: string | null;
+  /** Shortfall: engineVault - vaultTokenBalance (0n when ok). null when unevaluated. */
+  shortfall: string | null;
   checkedAt: number;
   skippedReason?: string;
 }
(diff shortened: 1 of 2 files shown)


dcccrypto:
Closing as **superseded** — `#394` (`0817b66`) implemented this independently.

`src/services/monitor.ts` on `main` now carries the fix this PR was for, with the reasoning spelled out in the same terms:

```
#347: `null` exists because this used to report `true` for v17 markets
without checking anything. Since the v17 cutover that is every production
market, so the only fund-conservation tripwire in the codebase was emitting
...
"UNCHECKED — v17 market account: the legacy engine.vault invariant does not apply
 and no v17 equivalent is implemented yet (#347). This is NOT a passing check."
```

Verified rather than assumed: I merged this branch into current `main` and diffed the result — **net effect zero**, no files changed. It is a literal no-op now.

Closing rather than leaving a no-op PR in the queue. #390 and #391 stay open — both still apply (`#390` +195 lines, and the CI docker job on `main` is still gated `if: github.ref == 'refs/heads/main'`, which is exactly what `#391` fixes). Full keeper suite with both merged: **1036 passed, 0 failed**.

PR #397, opened 2026-08-06
fix(keeper-send): broadcast the budget-gated priority fee + CU (closes #396)

### LINKED ISSUE #396: keeper: budget gate is decoupled from the priority fee actually broadcast (unbounded RPC-controlled spend)
## Summary

`keeperSend()` gates the SOL budget on a priority-fee estimate it computes locally, but never passes that estimate to the transaction it actually broadcasts. The shared sender (`sendWithRetryKeeper`) consequently derives the priority fee itself from `getRecentPrioritizationFees` — an RPC-controlled value with a floor but no ceiling. The fee the budget approved and the fee bid on-chain are independent numbers drawn from two different sources.

**Severity: Critical** — unbounded, RPC-influenced keeper spend that bypasses the budget gate.

## Impact

A malicious or degraded RPC can answer the keeper's fee-estimator call (`getPriorityFeeEstimate`, tier percentile) with a small value, so the budget gate approves a low `estimatedCost`, while answering the distinct `getRecentPrioritizationFees` call with an arbitrarily large p75. The keeper then signs and broadcasts `setComputeUnitPrice(p75) × setComputeUnitLimit(limit)` — a priority bid bounded only by wallet balance — having reserved only the small estimate against the budget.

Even with no adversary, the two paths diverge structurally: the budget gates on `simulatedCu × heliusTierFee` while the chain charges `computeUnitLimit × getRecentPrioritizationFees`. The realized-cost reconciliation books the true fee only ~5s after confirmation, so the spend has already happened by the time the budget learns of it.

## Root cause

`src/lib/keeper-send.ts` — `estimateCost()` computes `microLamports` (the tier-aware priority fee) and `simulatedCu`, gates the budget on both, then builds the send options with neither:

```ts
const opts: KeeperSendOptions = {
  ...keeperOpts,
  ...(isMainnetSender() ? { skipPreflight: true } : {}),
};
const signature = await sendWithRetryKeeper(connection, instructions, signers, maxRetries, opts);
```

`@percolatorct/shared` `sendWithRetryKeeper()` (standard, non-Sender path) then derives its own fee because the override is absent:

```js
const priorityFeeMicroLamports = keeperOpts?.priorityFeeMicroLamports ??
    (await getRecentPriorityFees(connection)).priorityFeeMicroLamports; // floor-only: Math.max(fee, 1_000), no ceiling
...
tx.add(ComputeBudgetProgram.setComputeUnitPrice({ microLamports: priorityFeeMicroLamports }));
```

The `priorityFeeMicroLamports` / `computeUnitLimit` overrides on `KeeperSendOptions` exist precisely to close this gap — the shared JSDoc states they are honored "so the budget the caller gated on is the one actually broadcast" (#311) — but no keeper call site populates them (verified: zero `src/` references to either field).

## Proof of Concept

`tests/lib/keeper-send.priority-fee-coupling.poc.test.ts`. It mocks the shared sender to capture the options `keeperSend` hands it, then models the shared sender's fee selection (`keeperOpts?.priorityFeeMicroLamports ?? getRecentPrioritizationFees()`) against a hostile RPC. Against current `main` it fails — the broadcast fee resolves to the RPC-controlled value and `keeperSend` passes no override:

```
expected 1000000000 to be 1000   // broadcast fee vs. gated fee
expected undefined  to be 1000   // keeperSend passes no priorityFeeMicroLamports
```

```ts
import { describe, it, expect, vi, beforeEach } from "vitest";

// What the keeper's estimator returns — the value the budget gate approves.
const GATED_FEE_MICROLAMPORTS = 1_000;
const GATED_CU = 200_000;
// What a hostile/degraded RPC returns from getRecentPrioritizationFees — the
// value the shared sender broadcasts when the keeper passes no override.
const HOSTILE_RPC_P75_MICROLAMPORTS = 1_000_000_000;

// Capture the keeperOpts (5th arg) keeperSend hands the shared sender.
const captured: { opts: any } = { opts: undefined };

  createLogger: vi.fn(() => ({ info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() })),
  sendWithRetryKeeper: vi.fn(async (_conn, _ix, _signers, _maxRetries, keeperOpts) => {
    captured.opts = keeperOpts;
    return "mock-sig";
  }),
}));

vi.mock("../../src/lib/priority-fee.js", () => {
  class HeliusPriorityFeeEstimator {
    estimate = vi.fn(async () => GATED_FEE_MICROLAMPORTS);
  }
  return { HeliusPriorityFeeEstimator };
});

vi.mock("../../src/lib/cu-estimator.js", () => {
  class CuEstimator {
    estimate = vi.fn(async () => ({ cu: GATED_CU, provenToFail: false }));
  }
  return { CuEstimator };
});

import { keeperSend } from "../../src/lib/keeper-send.js";
import { KeeperBudget } from "../../src/lib/budget.js";
import { Keypair, TransactionInstruction, PublicKey } from "@solana/web3.js";

function makeDummyIx(): TransactionInstruction {
  return new TransactionInstruction({ programId: PublicKey.default, keys: [], data: Buffer.from([]) });
}

function makeConnection() {
  return {
    simulateTransaction: vi.fn(async () => ({ value: { unitsConsumed: GATED_CU, err: null, logs: [] } })),
    getTransaction: vi.fn(async () => ({ meta: { fee: 5_000, err: null } })),
  } as any;
}

describe("PoC: budget-gated priority fee must equal the broadcast priority fee", () => {
  beforeEach(() => {
    vi.clearAllMocks();
    captured.opts = undefined;
    process.env.NETWORK = "devnet";
    process.env.USE_HELIUS_SENDER = "false";
    process.env.KEEPER_REALIZED_COST_SAMPLE_PCT = "0";
    delete process.env.PRIORITY_FEE_MICROLAMPORTS;
    delete process.env.DRY_RUN;
  });

  it("plumbs the gated priority fee + CU into the shared sender (no RPC-controlled bid)", async () => {
    const budget = new KeeperBudget({ maxSolPerCycle: 1_000_000_000 });
    const result = await keeperSend(makeConnection(), [makeDummyIx()], [Keypair.generate()], "liquidation", budget, 3);

    expect(result).not.toBeNull();
    expect(captured.opts).toBeDefined();

    // Shared sender selection (verified in @percolatorct/shared):
    //   fee = keeperOpts?.priorityFeeMicroLamports ?? getRecentPrioritizationFees()
    //   cu  = keeperOpts?.computeUnitLimit          ?? 400_000
    const broadcastFee = captured.opts.priorityFeeMicroLamports ?? HOSTILE_RPC_P75_MICROLAMPORTS;
    const broadcastCu = captured.opts.computeUnitLimit ?? 400_000;

    expect(broadcastFee).toBe(GATED_FEE_MICROLAMPORTS);
    expect(broadcastCu).toBe(GATED_CU);
  });

  it("a caller-supplied override still wins (does not clobber an explicit fee)", async () => {
    const budget = new KeeperBudget({ maxSolPerCycle: 1_000_000_000 });
    const result = await keeperSend(
      makeConnection(), [makeDummyIx()], [Keypair.generate()], "crank", budget, 3,
      { priorityFeeMicroLamports: 7_777, computeUnitLimit: 123_456 },
    );
    expect(result).not.toBeNull();
    expect(captured.opts?.priorityFeeMicroLamports).toBe(7_777);
    expect(captured.opts?.computeUnitLimit).toBe(123_456);
  });
});
```

## Suggested fix

Plumb the already-computed estimate into the send options so the gate and the broadcast use a single number:

```ts
priorityFeeMicroLamports: keeperOpts?.priorityFeeMicroLamports ?? microLamports,
computeUnitLimit:         keeperOpts?.computeUnitLimit ?? simulatedCu,
```

This makes the budget gate the effective ceiling: an adversarial or spiking fee now inflates `estimatedCost` and trips the existing fail-closed budget halt instead of overspending, and it aligns the CU basis the budget reserves against with the CU actually requested on-chain.

## Residual (separate, mainnet-only)

The Helius Sender fast path (`USE_HELIUS_SENDER=true`) forwards `computeUnitLimit` but not `priorityFeeMicroLamports` to `sendKeeperTxViaSender`, deriving the fee via `getHeliusPriorityFee` (also floor-only). Fully closing the Sender path requires a change in `@percolatorct/shared`; the standard path fix above covers devnet, which does not use the Sender.

0x-SquidSol:
Closes #396.

## Problem

`keeperSend()` gates the SOL budget on a priority-fee estimate it computes locally (`estimateCost` → `HeliusPriorityFeeEstimator`, tier percentile), then builds the `KeeperSendOptions` handed to `sendWithRetryKeeper` with neither `priorityFeeMicroLamports` nor `computeUnitLimit`. On the standard (non-Sender) path, `@percolatorct/shared` therefore re-derives the priority fee from `getRecentPrioritizationFees` — a distinct, RPC-controlled call with a floor (`Math.max(fee, 1_000)`) but no ceiling — and broadcasts it via `setComputeUnitPrice`.

The fee the budget approved and the fee bid on-chain were independent numbers from different sources. A malicious or degraded RPC could return a small value to the estimator (budget approves a low `estimatedCost`) and an arbitrarily large p75 to `getRecentPrioritizationFees` (broadcast bids that large), spending past the budget cap bounded only by wallet balance.

## Fix

Surface `microLamports` from `estimateCost` and pass it — together with `simulatedCu` — as the send overrides `@percolatorct/shared` already exposes for exactly this purpose (#311, "so the budget the caller gated on is the one actually broadcast"):

```ts
priorityFeeMicroLamports: keeperOpts?.priorityFeeMicroLamports ?? microLamports,
computeUnitLimit:         keeperOpts?.computeUnitLimit ?? simulatedCu,
```

Now the gate and the broadcast use one number. An inflated fee raises `estimatedCost` and trips the existing fail-closed budget halt instead of overspending, and the CU basis the budget reserves against matches the CU requested on-chain. Any caller-supplied override still takes precedence.

**Second-order effect: this also lowers real spend on every crank.** Solana charges the priority fee on the *requested* CU limit, not the amount consumed. The crank call sites pass `simulateForCU: false`, so with no `computeUnitLimit` override the shared sender fell through to a flat `400_000` — a crank consuming ~50k CU was bidding on 400k. Passing `simulatedCu` (`Math.ceil(consumed * 1.1)`, safe headroom) cuts the priority component proportionally.

## Testing

- New PoC `tests/lib/keeper-send.priority-fee-coupling.poc.test.ts` captures the options `keeperSend` hands the sender and asserts the broadcast fee/CU equal the gated estimate (fails before this change: broadcast resolves to the RPC value / `undefined`; passes after). Includes a case proving a caller-supplied override is not clobbered.
- Full suite green (`1035 passed | 33 skipped`); `tsc --noEmit` clean.

## Residual (out of scope, mainnet-only) — this PR does not close the whole class

The Helius Sender fast path (`USE_HELIUS_SENDER=true`) forwards `computeUnitLimit` but not `priorityFeeMicroLamports` to `sendKeeperTxViaSender`, deriving the fee via `getHeliusPriorityFee` (also floor-only), so it self-derives regardless of this change. Closing it requires a `@percolatorct/shared` fix and will land as part of a separate, dedicated Sender-routing change — closing #396 here is scoped to the standard path only (which is what devnet uses; the Sender path is mainnet-only).

### DIFF
--- a/src/lib/keeper-send.ts
+++ b/src/lib/keeper-send.ts
@@ -119,6 +119,13 @@ function firstProgramInstruction(instructions: TransactionInstruction[]): Transa
 interface EstimateCostResult {
   estimatedCost: number;
   simulatedCu: number;
+  /**
+   * The priority fee (microLamports per CU) the budget gated on. Surfaced so the
+   * caller can broadcast this exact value instead of letting the shared sender
+   * re-derive it from getRecentPrioritizationFees — otherwise the gated cost and
+   * the on-chain bid are independent numbers (dcccrypto/percolator-keeper#396).
+   */
+  microLamports: number;
   /** H-3: true iff simulation proved this exact tx would fail on-chain. */
   provenToFail: boolean;
   simError?: unknown;
@@ -153,6 +160,7 @@ async function estimateCost(
   return {
     estimatedCost: estimateLamportCost(microLamports, cuResult.cu, jitoTip, extraLamports),
     simulatedCu: cuResult.cu,
+    microLamports,
     provenToFail: cuResult.provenToFail,
     simError: cuResult.simError,
   };
@@ -234,7 +242,7 @@ export async function keeperSend(
     ...instructions,
   ];
 
-  const { estimatedCost, simulatedCu, provenToFail, simError } = await estimateCost(
+  const { estimatedCost, simulatedCu, microLamports, provenToFail, simError } = await estimateCost(
     connection,
     simulationInstructions,
     signers,
@@ -336,6 +344,17 @@ export async function keeperSend(
 
     const opts: KeeperSendOptions = {
       ...keeperOpts,
+      // Couple the broadcast to the budget: hand the shared sender the exact
+      // priority fee and CU limit the budget gated on, so it broadcasts those
+      // instead of re-deriving the fee from getRecentPrioritizationFees (an
+      // RPC-controlled, ceiling-less source). Without this the gated cost and the
+      // on-chain bid are independent numbers, and a hostile or spiking RPC can
+      // drive spend past the budget cap while the gate approved a small estimate;
+      // with it, an inflated fee raises estimatedCost and trips the existing
+      // fail-closed budget halt instead. A caller-supplied override still wins.
+      // See dcccrypto/percolator-keeper#396.
+      priorityFeeMicroLamports: keeperOpts?.priorityFeeMicroLamports ?? microLamports,
+      computeUnitLimit: keeperOpts?.computeUnitLimit ?? simulatedCu,
       // Saves ~20-50ms on mainnet when Helius Sender runs its own preflight downstream.
       ...(isMainnetSender() ? { skipPreflight: true } : {}),
     };
--- a/tests/lib/keeper-send.priority-fee-coupling.poc.test.ts
+++ b/tests/lib/keeper-send.priority-fee-coupling.poc.test.ts
@@ -1,0 +1,165 @@
+/**
+ * PoC — CRITICAL: budget gate is decoupled from the priority fee actually broadcast.
+ *
+ * keeperSend computes a tier-aware priority-fee estimate (HeliusPriorityFeeEstimator)
+ * and a CU estimate, then gates the budget on
+ *     estimatedCost = base + microLamports * simulatedCu / 1e6
+ * ...and hands `sendWithRetryKeeper` an options object containing NEITHER
+ * `priorityFeeMicroLamports` NOR `computeUnitLimit`.
+ *
+ * Per @percolatorct/shared (#311), when those overrides are absent the shared
+ * sender derives the broadcast fee ITSELF from `getRecentPrioritizationFees` — a
+ * DIFFERENT, RPC-controlled source — with a floor-only `Math.max(fee, 1_000)` and
+ * NO ceiling, then broadcasts it via setComputeUnitPrice(microLamports).
+ *
+ * Exploit: a malicious or merely spiking RPC answers the keeper's estimator call
+ * with a small fee (budget approves ~5_200 lamports) and answers
+ * getRecentPrioritizationFees with an arbitrarily large p75. The keeper signs and
+ * broadcasts a priority bid bounded only by wallet balance, while the budget
+ * booked the small estimate. The override that prevents exactly this already
+ * exists in shared — its JSDoc says "so the budget the caller gated on is the one
+ * actually broadcast" — the keeper simply never populates it.
+ *
+ * This PoC captures the options keeperSend hands the (mocked) shared sender and
+ * asserts the broadcast fee + CU equal the budget-gated estimate. It FAILS before
+ * the fix (options undefined → shared would broadcast the RPC's fee) and PASSES
+ * after keeperSend plumbs its gated estimate through.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// What the keeper's estimator returns — the value the budget gate approves.
+const GATED_FEE_MICROLAMPORTS = 1_000;
+const GATED_CU = 200_000;
+// What a hostile/degraded RPC returns from getRecentPrioritizationFees — the
+// value the shared sender broadcasts when the keeper passes no override.
+const HOSTILE_RPC_P75_MICROLAMPORTS = 1_000_000_000;
+
+// Capture the keeperOpts (5th arg) keeperSend hands the shared sender.
+const captured: { opts: any } = { opts: undefined };
+
+vi.mock("@percolatorct/shared", () => ({
+  createLogger: vi.fn(() => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  })),
+  sendWithRetryKeeper: vi.fn(async (_conn, _ix, _signers, _maxRetries, keeperOpts) => {
+    captured.opts = keeperOpts;
+    return "mock-sig";
+  }),
+}));
+
+vi.mock("../../src/lib/priority-fee.js", () => {
+  class HeliusPriorityFeeEstimator {
+    estimate = vi.fn(async () => GATED_FEE_MICROLAMPORTS);
+  }
+  return { HeliusPriorityFeeEstimator };
+});
+
+vi.mock("../../src/lib/cu-estimator.js", () => {
+  class CuEstimator {
+    estimate = vi.fn(async () => ({ cu: GATED_CU, provenToFail: false }));
+  }
+  return { CuEstimator };
+});
+
+import { keeperSend } from "../../src/lib/keeper-send.js";
+import { KeeperBudget } from "../../src/lib/budget.js";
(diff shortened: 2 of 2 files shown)


dcccrypto (commented):
**Correct, and the test binds it** — 2 of 3 fail when I delete the two new option
lines. Keeper suite on this branch: `1035 passed / 0 failed` (99 files).

## The load-bearing claim verified against the shipped shared package

The whole fix rests on `sendWithRetryKeeper` actually honouring the overrides. If
it ignored them this would be a no-op that reads like a fix, so I checked the
installed artifact rather than the JSDoc —
`@percolatorct/shared/dist/utils/solana.js:302`:

```js
const priorityFeeMicroLamports = keeperOpts?.priorityFeeMicroLamports ??
    (await getRecentPriorityFees(connection)).priorityFeeMicroLamports;
```

Honoured. Same for `computeUnitLimit` at `:304-313`. The `??` also means a
caller-supplied override still wins, which the PR preserves correctly.

## The CU concern I chased — and it resolves in the PR's favour, twice

Pinning `computeUnitLimit` to a *simulated* value is the obvious way to break
this: too tight and every send fails on-chain with an exceeded-CU error. It isn't
tight — `cu-estimator.ts:97-100` returns `Math.ceil(consumed * this._margin)` with
`DEFAULT_MARGIN = 1.1`, tunable via `KEEPER_CU_SIMULATE_MARGIN`. So `simulatedCu`
already carries 10% headroom and is a safe limit.

**Second-order effect the PR undersells: this should also reduce real spend.**
Solana charges the priority fee on the *requested* CU limit, not the consumed
amount. Previously, with no `computeUnitLimit` override and `simulateForCU: false`
at the crank call sites (`crank.ts:494`, `:624`, `:661`), the shared sender fell
through to:

```js
else { computeUnitLimit = 400_000; }
```

So a crank consuming ~50k CU was bidding on 400k. Requesting `simulatedCu`
instead cuts the priority component proportionally. Worth stating in the PR body —
"couples the gate to the broadcast" undersells a change that also lowers the bill
on every crank. (`fee-crank.ts:761` already passed `simulateForCU: true`, so that
path was closer to right.)

## Severity — I agree it's real, with one framing note

The adversarial-RPC path is genuine: two *different* RPC calls
(`getPriorityFeeEstimate` vs `getRecentPrioritizationFees`) feeding the gate and
the bid means a hostile endpoint can answer them inconsistently, and only the
second reaches the chain.

But the non-adversarial half is the more compelling argument and deserves top
billing: the two paths **diverge structurally even with an honest RPC** — tier
percentile vs p75 of the last 150 slots, `simulatedCu` vs a flat 400k — and the
realized-cost reconciliation only books the truth ~5s post-confirmation. A budget
that reserves one number and spends another is broken regardless of adversary,
and that framing doesn't depend on anyone accepting a malicious-RPC threat model.

## Residual, as scoped

Agreed the Helius Sender path needs a `@percolatorct/shared` change — `:280`
forwards `computeUnitLimit` to `sendKeeperTxViaSender` but not
`priorityFeeMicroLamports`, so that path still self-derives. Correctly called out
as out of scope; devnet doesn't use Sender. Worth its own issue against `shared`
so it isn't lost — this PR closing #396 might read as closing the whole class.

PR #399, opened 2026-08-06
fix(leader): demote by lease deadline when renewal stalls (closes #398)

### LINKED ISSUE #398: leader lock: renewal has no timeout — a hung Redis keeps a node leader past lease expiry (split-brain)
## Summary

`LeaderLock._renew()` awaits `redis.eval(RENEW_SCRIPT, ...)` with no timeout, and the Upstash client is constructed with no request signal (`new Redis({ url, token })`). Demotion is purely event-driven — it happens only when the renew eval **throws** (the 2-strike counter) or **returns** a lease-lost result. A hung or partitioned Redis whose request neither resolves nor rejects produces neither, so `_renew` stalls forever, no new renew timer is scheduled, and `_role` stays `"leader"` indefinitely while the Redis lock TTL lapses and a standby acquires it.

**Severity: High** — split-brain. Two nodes act as leader against a permissionless on-chain program, so the host-local single-writer barrier that guards against duplicate liquidations/sends is defeated exactly when it is needed (a Redis incident).

## Impact

The keeper's send path gates on `leaderLock.role()`. While one node is stuck believing it is leader (hung renewal), a standby — on the other side of the partition, or after Redis recovers for it — sees the expired lock and acquires it. Both nodes then submit liquidations/cranks. A secondary facet: even when the eval *does* eventually throw, the 2-strike counter means demotion takes up to `2 × (call latency) + renewMs`; with unbounded call latency this can exceed `ttlMs`, leaving a dual-leader window. The existing code acknowledges the tolerance is provisional (`leader.ts`: "this is the path the M15 finding discusses; tightening it is a separate PR").

## Root cause

`src/lib/leader.ts` — `_renew()`:

```ts
private async _renew(opts: StartOptions): Promise<void> {
  if (this._stopped || this._role !== "leader") return;
  try {
    const result = await this.redis.eval<number | null>(   // no timeout; can hang forever
      RENEW_SCRIPT, [this._lockKey], [this.identity, this.ttlMs],
    );
    if (result === 1) { this._renewFailures = 0; this._scheduleRenew(opts); return; }
    // ... lease-lost / strike handling only runs if eval resolves or throws ...
  } catch (err) { /* 2-strike demote */ }
}
```

`src/lib/redis-client.ts` — `new Redis({ url, token })` sets no `signal`/timeout, so a hung underlying fetch stalls the await indefinitely.

## Proof of Concept

`tests/lib/leader-hung-renew.poc.test.ts`. Every renewal hangs; the test asserts the node has demoted itself by the lease deadline. Against current `main` it fails — the node is still `"leader"` 31s in (`Expected "standby", Received "leader"`):

```ts
import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
import { LeaderLock } from "../../src/lib/leader.js";
import type { RedisLike } from "../../src/lib/redis-client.js";

const KEY = "keeper:leader:devnet";
type SetOpts = { ex: number; nx?: true } | { ex: number; xx?: true };

describe("LeaderLock hung-renewal lease expiry (PoC)", () => {
  beforeEach(() => { vi.useFakeTimers(); });
  afterEach(() => { vi.useRealTimers(); vi.restoreAllMocks(); });

  it("demotes by the lease deadline when Redis renewal hangs", async () => {
    const store = new Map<string, string>();
    const redis: RedisLike = {
      async set(key: string, value: string, opts: SetOpts): Promise<"OK" | null> {
        const hasNx = "nx" in opts && opts.nx === true;
        if (hasNx && store.has(key)) return null;
        store.set(key, value);
        return "OK";
      },
      async get(key: string): Promise<string | null> { return store.get(key) ?? null; },
      async del(...keys: string[]): Promise<number> {
        let n = 0; for (const k of keys) if (store.delete(k)) n++; return n;
      },
      // Renewal hangs forever — Redis partitioned/unresponsive mid-request.
      eval<T>(): Promise<T> { return new Promise<T>(() => {}); },
    };

    const lock = new LeaderLock(redis, "node-a", { ttlMs: 30_000, renewMs: 10_000, pollMs: 5_000 });
    const onDemote = vi.fn();
    lock.start({ network: "devnet", onPromote: vi.fn(), onDemote });

    await vi.advanceTimersByTimeAsync(100);
    expect(lock.role()).toBe("leader");
    expect(store.has(KEY)).toBe(true);

    // Renewal fires at 10s and hangs; by the 30s lease deadline the node must relinquish.
    await vi.advanceTimersByTimeAsync(31_000);

    expect(lock.role()).toBe("standby");
    expect(onDemote).toHaveBeenCalled();

    await lock.stop();
  });
});
```

## Suggested fix

Two complementary changes, both keeper-side in `leader.ts`:

1. **Lease-expiry watchdog.** Arm a timer for `ttlMs` on promotion and re-arm it on every successful renewal; if it fires while still leader, demote (`"lease-expired"`). This guarantees the node relinquishes leadership by the deadline of the lease it last set, regardless of why renewal stopped — the hard split-brain guarantee.
2. **Bound the renew eval with a timeout** (`Promise.race`), and re-check `_stopped`/`_role` after the await so a hung eval that resolves after a demotion cannot re-arm timers on a non-leader. This prevents the indefinite stall and keeps the strike path responsive.

0x-SquidSol:
Closes #398.

## Problem

`LeaderLock._renew()` awaited `redis.eval(RENEW_SCRIPT, ...)` with no timeout, and the Upstash client is constructed with no request signal. Demotion was purely event-driven — only when the eval **threw** (2-strike counter) or **returned** a lease-lost result. A hung or partitioned Redis whose request neither resolves nor rejects produced neither, so `_renew` stalled forever, no new renew timer was scheduled, and `_role` stayed `"leader"` indefinitely while the lock TTL lapsed and a standby acquired it — split-brain double-send against a permissionless program.

## Fix

Keeper-side, in `leader.ts`:

1. **Lease-expiry watchdog** — armed for `ttlMs` on promotion and re-armed on every successful renewal. If it fires while still leader, demote (`"lease-expired"`). This guarantees a node relinquishes leadership by the deadline of the lease it last set, regardless of why renewal stopped succeeding — the hard split-brain guarantee behind the existing 2-strike counter.
2. **Bounded renew eval** — `Promise.race` against `renewTimeoutMs` (default `min(renewMs, 5s)`), surfaced as a thrown error onto the existing strike/demote path, so a hung request can no longer stall the loop.
3. **Post-await role re-check** — after the eval and in the catch, bail if a concurrent `stop()`/watchdog demotion already moved us out of leader, so a late-resolving eval cannot re-arm timers on a non-leader.

## Testing

- New PoC `tests/lib/leader-hung-renew.poc.test.ts` hangs every renewal and asserts the node has demoted by the lease deadline (fails before: stuck `"leader"` at 31s; passes after).
- All existing leader/C2 tests unchanged and green (`leader.test.ts`, `leader-demote-recovery.poc.test.ts`, `keeper-send-leader-gate.test.ts`, `c2.poc.test.ts`); full suite `1033 passed`; `tsc --noEmit` clean.

## Summary by CodeRabbit

* **New Features**
  * Added configurable renewal timeouts and lease monitoring for leader locks.
  * Added optional persistence and restoration of keeper budget halt states.
  * Updated LP crank operations to support dual-domain backing ledgers and lazy ledger creation.

* **Bug Fixes**
  * Prevented stalled or delayed renewals from incorrectly restoring leadership.
  * Mainnet startup now requires the approved transaction sender configuration.

* **Documentation**
  * Documented recommended lease timing requirements and safety margins.

* **Tests**
  * Added coverage for renewal hangs, lease validation, halt persistence, and startup safeguards.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -91,6 +91,13 @@ USE_HELIUS_SENDER=true
 # NETWORK=mainnet
 # KEEPER_REDIS_URL=https://your-upstash-rest-url.upstash.io
 # KEEPER_REDIS_TOKEN=                # required when KEEPER_REDIS_URL is set (no empty fallback)
+# Lease timings. All three must be finite and > 0, and RENEW_MS must be
+# strictly less than TTL_MS — otherwise the lease expires before the leader
+# renews it, a standby promotes while the old leader still believes it holds
+# the lock, and both submit on-chain transactions (split-brain). The keeper
+# refuses to boot on a violation rather than run split-brained (#377).
+# Keep RENEW_MS at or below half of TTL_MS so a slow Redis round-trip or an
+# event-loop stall cannot let the lease lapse between renewals.
 # KEEPER_LEADER_LOCK_TTL_MS=30000
 # KEEPER_LEADER_LOCK_RENEW_MS=10000
 # KEEPER_STANDBY_POLL_MS=5000
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,44 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     needs: build-and-test
-    if: github.ref == 'refs/heads/main'
+    # #384: this was gated to `refs/heads/main`, so the image was only built
+    # AFTER merge and image-build breakage passed PR review invisibly. It then
+    # failed on every push to main from 2026-06-22 to 2026-07-18 (~4 weeks) with
+    # ERR_PNPM_LOCKFILE_CONFIG_MISMATCH, and the PR that introduced it could not
+    # have caught it: build-and-test passes there, because actions/checkout
+    # supplies the whole repo while the docker build context is narrower. The
+    # two jobs disagree on exactly the thing that broke, and only the invisible
+    # one failed.
+    #
+    # The job only builds — it never pushes to a registry — so running it on
+    # pull requests needs no push-gating.
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Build Docker image
         run: docker build -t percolator-keeper .
+
+      # K-NEW-1 in the Dockerfile installs production deps only, to keep vitest,
+      # tsx, @types/* and their CVEs out of the runtime image. Nothing verified
+      # that. A regression to a full install would still build green, so assert
+      # the outcome rather than the exit code.
+      - name: Assert devDependencies are absent from the runtime image
+        run: |
+          set -euo pipefail
+          for pkg in vitest tsx typescript fast-check; do
+            if docker run --rm --entrypoint sh percolator-keeper -c "test -e /app/node_modules/$pkg"; then
+              echo "::error::devDependency '$pkg' is present in the runtime image (K-NEW-1 regression)"
+              exit 1
+            fi
+          done
+          echo "OK: vitest, tsx, typescript and fast-check are absent from the runtime image"
+
+      - name: Assert production dependencies survived
+        run: |
+          set -euo pipefail
+          # Mirror check: an install that dropped real dependencies would also
+          # satisfy the assertion above, so it has to be paired with this one.
+          for pkg in @solana/web3.js @upstash/redis p-queue lru-cache; do
+            docker run --rm --entrypoint sh percolator-keeper -c "test -e /app/node_modules/$pkg" \
+              || { echo "::error::production dependency '$pkg' missing from the runtime image"; exit 1; }
+          done
+          echo "OK: production dependencies are present"
--- a/.github/workflows/sdk-smoke.yml
+++ b/.github/workflows/sdk-smoke.yml
@@ -30,7 +30,7 @@ on:
 
 env:
   # Update this string when the SDK bumps.  Deliberate — that is the point.
-  SDK_VERSION: "2.0.5"
+  SDK_VERSION: "4.4.0"
 
 concurrency:
   group: sdk-smoke-${{ github.ref }}
@@ -38,7 +38,7 @@ concurrency:
 
 jobs:
   smoke:
-    name: smoke @percolatorct/sdk@${{ github.event.client_payload.version || inputs.sdk_version || '2.0.5' }}
+    name: smoke @percolatorct/sdk@${{ github.event.client_payload.version || inputs.sdk_version || '4.4.0' }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 .env
 *.tsbuildinfo
 .pnpm-store/
+.claude/
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
-    "@percolatorct/sdk": "github:dcccrypto/percolator-sdk#bbe48e445e8e6aae386d9bfb6c7ad5970f92a3f0",
+    "@percolatorct/sdk": "github:dcccrypto/percolator-sdk#886aa6b9eb96950c8e415df695cd3ae3614472d5",
     "@percolatorct/shared": "github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f",
     "@solana/web3.js": "^1.98.0",
     "@upstash/redis": "^1.38.0",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
-    "@percolatorct/sdk": "github:dcccrypto/percolator-sdk#ac68ac3f45fb3a069866308727c0fd40e8293d2e",
+    "@percolatorct/sdk": "github:dcccrypto/percolator-sdk#bbe48e445e8e6aae386d9bfb6c7ad5970f92a3f0",
     "@percolatorct/shared": "github:dcccrypto/percolator-shared#052edb592023de7c097f46edf8fba9a1cb20423f",
     "@solana/web3.js": "^1.98.0",
     "@upstash/redis": "^1.38.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   vite: 8.0.8
+  '@opentelemetry/core': '>=2.8.0'
 
 importers:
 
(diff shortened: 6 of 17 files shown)


dcccrypto (commented):
**Correct, and a genuinely careful fix.** Keeper suite on this branch:
`1033 passed / 0 failed`. Two things I want to flag: a residual window the fix
narrows but doesn't fully close, and what the PoC does and doesn't pin.

The defensive details are right and worth naming, because they're the parts that
usually get missed in a fix like this:

- **Watchdog armed before `onPromote()`**, so the split-brain guard is live even if
  the promote handler throws — matching `_demote`'s existing arm-poll-first
  ordering.
- **Stale-result guards in both the success and catch paths.** A hung eval that
  resolves *after* the watchdog already demoted us would otherwise re-arm timers on
  a non-leader. This is the subtle one and it's handled.
- **`_clearTimers` covers `_leaseTimer`**, and `_demote` early-returns on
  non-leader, so a late-firing watchdog is inert.

## Residual: the watchdog is armed on the renewal *response*, not the *request*

`_armLeaseWatchdog()` runs when `result === 1` — i.e. when the reply arrives. But
the lease Redis grants starts when Redis *executes* the eval. With the defaults
(`ttlMs` 30 000, `renewMs` 10 000, `renewTimeoutMs` 5 000), a renewal that takes
4.9s — slow but under the timeout, so it succeeds — extends the Redis-side lease to
`T+30s` while the watchdog is armed for `T+34.9s`.

That leaves up to `renewTimeoutMs` of dual-leader window: the lock is free from
`T+30s`, a standby polling every `pollMs` (5s) can take it, and this node still
believes it is leader for another ~4.9s. Narrower than the unbounded window the PR
closes, but the same failure mode — and a slow-but-not-hung Redis is a more common
incident than a fully hung one.

Cheap to close: stamp the deadline at *issue* time rather than resolve time.

```ts
const issuedAt = Date.now();
const result = await this._evalRenewWithTimeout();
if (this._stopped || this._role !== "leader") return;
if (result === 1) {
  this._renewFailures = 0;
  this._armLeaseWatchdog(this.ttlMs - (Date.now() - issuedAt)); // deadline from when Redis got it
  ...
```

Same argument applies to `_promote`: the lease begins at the successful `SET NX`,
not when `_promote` runs. That gap is negligible today, but the issue-time
convention makes both correct by construction rather than by being fast.

## The PoC pins the outcome, not either mechanism

Worth knowing before someone simplifies this later. I removed each half
independently:

| mutation | PoC result |
|---|---|
| watchdog arming removed (timeout kept) | **passes** — demotes via the 2-strike path |
| eval timeout removed (watchdog kept) | **passes** — demotes via the watchdog at ttl |
| **both removed** | **fails** (`Expected "standby", Received "leader"`) |

So the test asserts "demoted by the lease deadline", which is arguably the right
level — but neither mechanism is individually pinned. The risk is specific: a
future reader sees two overlapping guards, concludes the timeout makes the watchdog
redundant, deletes it, and the suite agrees. It isn't redundant — the watchdog is
what still bounds demotion if `renewTimeoutMs` is ever configured above `ttlMs`
(nothing currently prevents `new LeaderLock({ renewTimeoutMs: 60_000 })`), or if a
future renew path forgets to route through `_evalRenewWithTimeout`.

A second case with `renewTimeoutMs` set high enough that strikes can't demote in
time would pin the watchdog specifically, and doubles as a guard on that
configuration hazard.

dcccrypto:
Follow-up to my review, on coordination rather than the code — I should have
caught this in the first pass.

**#399 and my open #385 both rewrite `src/lib/leader.ts`, and they conflict.**
Test-merged onto `main` in sequence:

```
merge #399 → exit 0
merge #385 → exit 1   CONFLICT: src/lib/leader.ts
```

Not a criticism of either change — they're solving adjacent halves of the same
problem and were written independently. But whichever lands second needs a manual
rebase, so it's worth knowing now rather than at merge time.

**The overlap is also an opportunity.** #385 (`#377`) adds `validateTiming()` +
`LeaderLockTimingError`, rejecting lease timings that make split-brain
deterministic — `renewMs >= ttlMs`, NaN from a malformed env var, non-positive
values. Its argument is precisely the one behind this PR's watchdog: a
misconfiguration whose only symptom is two nodes writing on-chain at once.

That makes it the natural home for the configuration hazard I raised above.
`#385` predates `renewTimeoutMs` and so doesn't validate it (`grep` returns 0),
which leaves the gap I described: nothing rejects
`new LeaderLock({ renewTimeoutMs: 60_000, ttlMs: 30_000 })`, and in that
configuration the timeout can't demote before the lease expires — the watchdog is
the only thing left, which is exactly why it isn't redundant.

Suggested split, so neither PR grows:

- **#399** keeps the runtime guarantees (timeout + watchdog) as-is.
- **#385** adds `renewTimeoutMs` to `validateTiming` — `0 < renewTimeoutMs <= renewMs < ttlMs`
  — once #399 has introduced the option.

I'm happy to rebase #385 onto #399 and add that check as soon as this lands, so
the ordering is: this PR first, mine after. Say the word if you'd rather it went
the other way and I'll rebase in whichever direction suits.

0x-SquidSol:
Applied the issue-time deadline fix (`1c19381`): both `_promote` (via `_tryAcquire` / `_poll`) and `_renew` now stamp the moment the SET/eval is issued and arm the watchdog for `ttlMs` minus the elapsed round-trip. That closes the up-to-`renewTimeoutMs` window you flagged for a slow-but-not-hung Redis — the node can no longer believe it holds a lease that has already expired on Redis's clock. Leader/C2 suites green.

On the watchdog-specific PoC case and the `renewTimeoutMs` config hazard: I'd rather that land as the `validateTiming` guard in **#385** (`0 < renewTimeoutMs <= renewMs < ttlMs`) than as a test that pins the current defaults — validation removes the hazard by construction, whereas a test only documents it. Happy to add a watchdog-only test here too if you'd still like the mechanism individually pinned.

Ordering: **#399 first works for me** — go ahead and rebase #385 on top and fold the `renewTimeoutMs` bound into `validateTiming`. Thanks for the careful review.

dcccrypto (commented):
Re-reviewed `1c19381`. **The fix is right, and it covers more than I asked for.**

Verified each arming site now dates from issue time:

| site | stamp | armed with |
|---|---|---|
| `_tryAcquire` | `acquiredAt` before `SET NX` | `_promote(opts, acquiredAt)` |
| **`_poll`** | `acquiredAt` before its `SET NX` | `_promote(opts, acquiredAt)` |
| `_renew` | `issuedAt` before the eval | `_armLeaseWatchdog(issuedAt)` |

**`_poll` is the one I didn't name.** I only flagged `_renew` and mentioned
`_promote` in passing; the standby→leader re-acquisition path has exactly the same
exposure and you caught it. Worth saying, because a partial fix there would have
left the window open on precisely the path that runs after a Redis incident — the
moment it matters most.

`Math.max(0, this.ttlMs - (Date.now() - leaseIssuedAt))` handles the degenerate
case correctly: a round trip that exceeded `ttlMs` arms at 0, so the watchdog fires
on the next tick and demotes rather than passing a negative delay. The callback's
`_role !== "leader"` re-check keeps that safe against a concurrent demote.

Checked it can't misfire in a healthy system: with the defaults, a renew that takes
4.9s (the worst case that still succeeds) arms for ~25.1s while the next renew is
scheduled 10s out — no premature demotion. And the comment's framing is the right
one: measuring from issue time is **conservative — arms slightly early, never
late**.

Suite on this head: `1033 passed / 0 failed`.

## Two notes, neither blocking

**The PoC doesn't bind this change.** I reverted `remainingMs` to `this.ttlMs`
(i.e. back to response-time arming, reopening the window) and
`leader-hung-renew.poc.test.ts` stayed **1/1 green**. That's expected — it asserts
demotion by the deadline under a *fully hung* Redis, where issue-vs-response time
is indistinguishable. Binding this specific fix needs the slow-but-successful case:
a renew that resolves after ~`renewTimeoutMs - ε` and still reports success, then
asserting demotion by `issuedAt + ttlMs` rather than `respondedAt + ttlMs`. With
fake timers that's a short test, and it's the only thing that would catch a
future revert.

**The #385 conflict persists**, as expected — re-tested against the new head:
`#399` merges clean onto `main`, then `#385` conflicts on `src/lib/leader.ts`. My
offer stands: land this first and I'll rebase #385 onto it, folding
`renewTimeoutMs` into its `validateTiming` so the `renewTimeoutMs > ttlMs`
misconfiguration is rejected at construction.

PR #402, opened 2026-08-06
fix(keeper): match custom program error 0x4 exactly, not as a substring (closes #401)

### LINKED ISSUE #401: keeper: 0x4 substring match permanently skips healthy markets on any 0x4X program error
## Summary

The permanent-skip guards in the crank and liquidation services classify a failed transaction as InvalidSlabLen (custom program error code 4) with `errMsg.includes("custom program error: 0x4")`. Solana formats custom errors as unpadded lowercase hex, so this substring also matches `0x40`–`0x4f` (codes 64–79) and `0x400+` (1024+). A transient engine error numbered in any of those ranges is misread as code 4 and the **healthy market is permanently removed** from cranking / liquidation.

**Severity: High** — a single unrelated `0x4X` error silently takes a healthy market dark (no cranks, no liquidations) until keeper restart; on repeated occurrences the crank path escalates the skip cooldown to 24h.

## Root cause

`src/services/crank.ts` (permanent skip → `state.permanentlySkipped = true`, `isActive = false`):

```ts
if (errMsg.includes("custom program error: 0x4")) {
  state.permanentlySkipped = true;
  ...
  return "skipped";
}
```

`src/services/liquidation.ts` (permanent skip → `this.permanentlySkipped.add(...)`):

```ts
if (errMsg.includes("custom program error: 0x4")) {
  this.permanentlySkipped.add(slabAddress.toBase58());
  ...
  return null;
}
```

Any `custom program error: 0x40` … `0x4f` / `0x400+` string satisfies the check.

## Proof of Concept

`tests/lib/program-error.poc.test.ts`, against an exact-match helper `isCustomProgramError(errMsg, code)`. The "does NOT match higher codes" case fails against the current substring check and passes against the boundary-anchored helper:

```ts
import { describe, it, expect } from "vitest";
import { isCustomProgramError } from "../../src/lib/program-error.js";

describe("isCustomProgramError — exact custom-error-code matching", () => {
  it("matches the exact code 0x4", () => {
    expect(isCustomProgramError("custom program error: 0x4", 4)).toBe(true);
    expect(isCustomProgramError(
      "Transaction simulation failed: Error processing Instruction 0: custom program error: 0x4", 4,
    )).toBe(true);
    expect(isCustomProgramError("custom program error: 0x4 (InvalidSlabLen)", 4)).toBe(true);
  });

  it("does NOT match higher codes that share the 0x4 prefix", () => {
    for (const higher of ["0x40", "0x41", "0x4c", "0x4f", "0x400", "0x4a2"]) {
      expect(isCustomProgramError(`custom program error: ${higher}`, 4)).toBe(false);
    }
  });

  it("documents the defect this replaces: the old substring check misfired", () => {
    expect("custom program error: 0x40".includes("custom program error: 0x4")).toBe(true);
    expect(isCustomProgramError("custom program error: 0x40", 4)).toBe(false);
  });

  it("works for multi-digit / hex-letter codes too", () => {
    expect(isCustomProgramError("custom program error: 0x33", 0x33)).toBe(true);
    expect(isCustomProgramError("custom program error: 0x330", 0x33)).toBe(false);
  });
});
```

## Fix

Extract one tested helper `isCustomProgramError(errMsg, code)` that anchors the hex code with a trailing word boundary, and route both permanent-skip guards through it with `code = 0x4`. This fixes the over-match and removes the duplicated inline check. The pinned exact-`0x4` behavior (crank.test.ts / liquidation.test.ts) is preserved.

## Related (low, not fixed here)

`crank.ts` also uses `errMsg.includes("custom program error: 0x33")` for an InsufficientDexLiquidity **log line** (code 51). It has the same substring-collision shape (`0x330+`) but only affects a log message, not skip state — routing it through the same helper would be a tidy follow-up.

0x-SquidSol:
Closes #401.

## Problem

The permanent-skip guards in `crank.ts` and `liquidation.ts` classified a failure as InvalidSlabLen (custom program error code 4) with `errMsg.includes("custom program error: 0x4")`. Solana formats custom errors as unpadded lowercase hex, so that substring also matches `0x40`–`0x4f` (codes 64–79) and `0x400+`. A transient engine error in those ranges was misread as code 4 and the healthy market was **permanently removed** from cranking / liquidation until keeper restart (the crank path additionally escalates the skip cooldown toward 24h on repeats).

## Fix

Extract one tested helper and route both guards through it:

```ts
// src/lib/program-error.ts
export function isCustomProgramError(errMsg: string, code: number): boolean {
  const hex = code.toString(16);
  return new RegExp(`custom program error: 0x${hex}\\b`).test(errMsg);
}
```

`crank.ts` and `liquidation.ts` now call `isCustomProgramError(errMsg, 0x4)`. The trailing `\b` (all hex digits are word characters) means `0x4` matches but `0x40` / `0x4c` / `0x400` do not. This also removes the duplicated inline check.

## Testing

- New `tests/lib/program-error.poc.test.ts` pins exact-code matching and documents the old misfire (fails against the substring check, passes against the helper).
- The pinned exact-`0x4` permanent-skip behavior in `crank.test.ts` / `liquidation.test.ts` is preserved and green; full suite `1036 passed`; `tsc --noEmit` clean.

## Related (not in this change)

`crank.ts` uses the same substring shape for a `0x33` InsufficientDexLiquidity **log line** (code 51). Same collision class (`0x330+`) but log-only, no skip-state impact — a tidy follow-up would route it through the same helper.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of program errors during automated market maintenance.
  - Prevented unrelated error codes from being mistaken for known errors.
  - Reduced the risk of markets being incorrectly skipped during crank and liquidation operations.

- **Tests**
  - Added coverage for exact error-code matching, embedded messages, named errors, and multi-digit codes.

### DIFF
--- a/src/lib/program-error.ts
+++ b/src/lib/program-error.ts
@@ -1,0 +1,15 @@
+/**
+ * Match a Solana custom program error code exactly.
+ *
+ * Solana formats a custom program error as unpadded lowercase hex:
+ * "custom program error: 0x4" for code 4, "0x40" for 64, "0x4c" for 76, "0x400"
+ * for 1024, etc. A substring check for "custom program error: 0x4" therefore
+ * also matches 0x40–0x4f and 0x400+, misreading a transient engine error in
+ * those ranges as code 4. Anchoring the code with a trailing word boundary
+ * (\b, since all hex digits are word characters) means only the exact code
+ * matches — "0x4" hits, "0x40"/"0x4c"/"0x400" do not.
+ */
+export function isCustomProgramError(errMsg: string, code: number): boolean {
+  const hex = code.toString(16);
+  return new RegExp(`custom program error: 0x${hex}\\b`).test(errMsg);
+}
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -25,6 +25,7 @@ import {
 import { config, getConnection, getFallbackConnection, loadKeypair, eventBus, createLogger, sendCriticalAlert, getSupabase } from "@percolatorct/shared";
 import { OracleService } from "./oracle.js";
 import { resolveExternalOracleAccount } from "../lib/oracle-account.js";
+import { isCustomProgramError } from "../lib/program-error.js";
 import { recordAttempt, recordLanded, recordFailed } from "../lib/sender-metrics.js";
 import {
   txSentTotal,
@@ -1626,9 +1627,11 @@ export class CrankService {
         });
       }
 
-      // Detect NotInitialized (error 0x4) — permanently skip these markets
-      // PERC-381: Track skip count and timestamp for exponential cooldown on rediscovery
-      if (errMsg.includes("custom program error: 0x4")) {
+      // Detect InvalidSlabLen (error 0x4) — permanently skip these markets.
+      // PERC-381: Track skip count and timestamp for exponential cooldown on rediscovery.
+      // Exact-code match: a substring check would also catch 0x40–0x4f / 0x400+
+      // and permanently skip a healthy market on an unrelated engine error.
+      if (isCustomProgramError(errMsg, 0x4)) {
         state.permanentlySkipped = true;
         state.permanentlySkippedAt = Date.now();
         state.skipCount = (state.skipCount ?? 0) + 1;
--- a/src/services/crank.ts
+++ b/src/services/crank.ts
@@ -512,7 +512,9 @@ async function provisionKeeperPortfolio(
   } catch (provisionErr) {
     const errMsg = provisionErr instanceof Error ? provisionErr.message : String(provisionErr);
     // AlreadyInitialized: a concurrent provision beat us; re-query for the existing portfolio.
-    if (errMsg.includes("AlreadyInitialized") || errMsg.includes("custom program error: 0x0")) {
+    // Exact-code match on 0x0 — a substring check also caught 0x00–0x0f (codes 1–15),
+    // misrouting an unrelated failure into this re-query path.
+    if (errMsg.includes("AlreadyInitialized") || isCustomProgramError(errMsg, 0x0)) {
       logger.info("provisionKeeperPortfolio: AlreadyInitialized — re-querying for existing portfolio", {
         market: marketKeyBase58.slice(0, 8),
       });
--- a/src/services/liquidation.ts
+++ b/src/services/liquidation.ts
@@ -19,6 +19,7 @@ import {
 import { config, getConnection, loadKeypair, sendWithRetry, pollSignatureStatus, getRecentPriorityFees, checkTransactionSize, eventBus, createLogger, sendWarningAlert, sendCriticalAlert, acquireToken, getFallbackConnection, backoffMs, getErrorMessage } from "@percolatorct/shared";
 import { OracleService } from "./oracle.js";
 import { resolveExternalOracleAccount } from "../lib/oracle-account.js";
+import { isCustomProgramError } from "../lib/program-error.js";
 import { recordAttempt, recordLanded, recordFailed } from "../lib/sender-metrics.js";
 import {
   txSentTotal,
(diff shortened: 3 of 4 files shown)


dcccrypto (commented):
**Correct, and the test binds it** — reverting `isCustomProgramError` to the old
`includes()` fails 3 of 4. Keeper suite on this branch: `1036 passed / 0 failed`.

The `\b` anchor is the right mechanism and worth spelling out for the next reader:
hex digits are word characters, so `0x4\b` requires the next character to be a
non-word one (or end of string). `"…0x4"` and `"…0x4 (InvalidSlabLen)"` match;
`"0x40"`, `"0x4c"`, `"0x400"` do not. `code.toString(16)` only ever yields
`[0-9a-f]+`, so there is no regex-injection surface in the interpolation either.

## Three more instances of the same defect, all in `crank.ts`, none fixed here

I swept for the pattern rather than assuming the two reported sites were all of
it. The helper this PR introduces is exactly what each of these needs:

| site | check | also matches | consequence |
|---|---|---|---|
| `crank.ts:515` | `custom program error: 0x0` | **`0x00`–`0x0f` (codes 1–15)** and `0x000+` | **behavioural** |
| `crank.ts:1598` | `custom program error: 0x25` | `0x250`–`0x25f`, `0x2500+` | log only |
| `crank.ts:1609` | `custom program error: 0x33` | `0x330`–`0x33f`, `0x3300+` | log only |

Calibrating honestly, because they are not equal:

- **`:515` is the one that matters.** `0x0` is the broadest possible prefix — it
  matches *every* code 1–15. On a false match the code takes the
  "AlreadyInitialized → re-query for the existing portfolio" path after an
  unrelated failure, and then hunts for a portfolio that was never created. Not as
  destructive as the permanent skip this PR fixes, but it is misrouted control
  flow, not just a bad log line.
- **`:1598` and `:1609` are diagnostic only** — both branches just call
  `logger.error`. A false match yields a misleading log entry and nothing more,
  and `:1609`'s own comment notes it is dead in v17. Worth fixing for consistency
  once the helper exists; not worth a separate PR on their own.

Also in the same expressions: the paired JSON-form checks have the same shape —
`includes("Custom\":37")` matches `"Custom":370` too. Same class, same fix, if the
helper ever grows a JSON-form branch.

`fee-crank.ts:192` already does this correctly:

```ts
const hex = /custom program error:\s*0x([0-9a-f]+)/i.exec(msg);
```

so the "extract the whole code, don't prefix-match" idiom was already in the repo —
which is a decent argument for `isCustomProgramError` becoming the single place
this logic lives, and for `fee-crank` eventually calling it too.

None of this blocks the PR: it fixes the high-severity instance (permanent skip of
a healthy market) and adds the helper the rest need. Folding `:515` in here would
be cheap and I would; `:1598`/`:1609` are fine as a follow-up.

dcccrypto:
The red `build-and-test` here is **GitHub Actions infrastructure, not this change.**

It never reaches your code — it fails during `Set up job`, resolving the action
definitions:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
##[error]Failed to resolve action download info.
```

I re-ran the failed job to check whether it was a one-off. It failed **identically
an hour later** (15:18 and 16:19), so this is a sustained degradation on GitHub's
side rather than a flake that clears on retry — a third attempt now is unlikely to
help, and the branch head is unchanged (`dc621ca`) between both runs.

For what it's worth, the suite is green on this exact commit locally:

```
Test Files  99 passed | 2 skipped (101)
     Tests  1036 passed | 33 skipped (1069)
```

Nothing to do on the PR. Worth knowing more broadly: while this persists, **any**
run in these repos will fail the same way at `Set up job`, so new red checks today
should be read at the failing step before they're attributed to a change.

0x-SquidSol:
Folded in the behavioural instance you flagged — `crank.ts` `provisionKeeperPortfolio` now uses `isCustomProgramError(errMsg, 0x0)` (`85eecca`), so an unrelated `0x01`–`0x0f` failure no longer misroutes into the AlreadyInitialized re-query path. The `0x25` / `0x33` log-only sites I've left as the follow-up you described.

Also rebased onto current `main` so this sits on top of the dual-domain LP-vault change (tag 78). `tsc --noEmit` clean and full suite green (`1036 passed`) against the re-pinned SDK (`886aa6b`). And agreed on the red check being the GitHub Actions `Set up job` outage, not this branch.

dcccrypto:
Re-reviewed `85eecca`. **`crank.ts:517` is folded in** — `0x0` now goes through
`isCustomProgramError`, closing the behavioural instance (that check matched every
code 1–15, misrouting unrelated failures into the "AlreadyInitialized → re-query"
path).

`:1612` (`0x25`) and `:1623` (`0x33`) are correctly left as substring checks. That
matches what I suggested: both branches only `logger.error`, so a false match costs
a misleading log line and nothing more, and `:1623`'s own comment notes it is dead
in v17. Fixing them here would have widened the diff for no behavioural gain.

Suite on this head, rebased onto the new `main` (`0ab802f`): `1036 passed / 0
failed`.

Nothing further from me.

Issue #372, opened 2026-06-27
Medium: bigint-buffer buffer-overflow CVE (GHSA-3gc7-fjrx-p6mg) has no patched version available — reachable on account-decoding path


Morenikeoa:
## Summary

\`pnpm audit\` flags \`bigint-buffer\` for a buffer-overflow vulnerability in \`toBigIntLE()\` (GHSA-3gc7-fjrx-p6mg), reachable transitively via \`@solana/spl-token\` -> \`@solana/buffer-layout-utils\` -> \`bigint-buffer\`, used on the keeper's core account-decoding path (\`crank.ts\`, \`liquidation.ts\`, \`rpc-pool.ts\` all decode SPL/account data fetched via \`getAccountInfo\`). Unlike the other dependency CVEs found in this audit round (e.g. \`ws\`), **no patched version of bigint-buffer has been published** as of this writing -- the latest available release (1.1.5) is the same version flagged as vulnerable.

## Environment

- Service: percolator-keeper (all three services touch account decoding)
- Version: main (post-v17)
- Advisory: https://github.com/advisories/GHSA-3gc7-fjrx-p6mg
- Vulnerable versions: \`<=1.1.5\`
- Patched versions: \`>=1.1.6\` (per the advisory) -- **but npm's registry currently has no version beyond 1.1.5 published** (\`npm view bigint-buffer versions\` tops out at \`1.1.5\`).

## Root Cause

\`toBigIntLE()\`/\`toBufferLE()\` in \`bigint-buffer\` have a buffer overflow on malformed/short input. Account data ultimately originates from RPC responses, which an adversarial or compromised RPC node could shape to trigger this.

## Production Impact

A malicious or compromised RPC endpoint could return crafted account data that triggers the overflow during decoding, on a process holding a hot signing wallet. Severity is bounded by the fact this requires either a compromised/malicious RPC provider or a MITM on an insecure RPC connection (already guarded against elsewhere in \`env-guards.ts\` via the HTTPS/WSS requirements).

## Reproduction Steps

1. Run \`pnpm audit\` against the current lockfile.
2. Observe the \`high\` severity \`bigint-buffer\` advisory with \`Patched versions: >=1.1.6\`.
3. Run \`npm view bigint-buffer versions --json\` and observe the highest published version is \`1.1.5\` -- there is currently no version satisfying \`>=1.1.6\` to upgrade or override to.

## Expected Behavior

[...]

PR #406, opened 2026-08-09
ci(security): add a dependency-audit gate and unpin vite from a vulnerable build (#372)

dcccrypto:
Closes the second half of #372.

## Why

The keeper had **no `pnpm audit` step in CI at all**. Every dependency CVE fixed in this repo so far — #378's `@opentelemetry/core`, the `ws` advisories — was found by hand. A newly published advisory against something we already ship could sit here indefinitely and nothing would say so. This is the service that holds a hot signing wallet, and it was the last of the four repos without a gate.

launch has the same gap tracked as launch#2502 (its gate exists but never runs); indexer and api already gate.

## What turning the gate on found

**11 high advisories — one of which this repo was causing itself.**

The existing override read `vite: 8.0.8`. GHSA-fx2h-pf6j-xcff covers `>=8.0.0 <=8.0.15`, so that exact pin — added before the advisory published — was holding vite *below* the fix. The override was the reason the advisory applied, not a defence against it. That is the argument for the gate in one line: a pin is a snapshot of what was safe on the day it was written, and nothing was re-checking it.

## The pin is still exact, deliberately

A `>=8.0.16` floor does **not** work here, and I want this recorded because it is counter-intuitive. A range override constrains what may be *selected*; it does not force an already-locked version to move. With `>=8.0.16` in the config and `8.0.8` in the lockfile, pnpm reports `Lockfile is up to date, resolution step is skipped` and leaves 8.0.8 in place. `pnpm install --force` and `pnpm update vite` do not shift it either, and you end up with a lockfile whose `overrides:` header says `>=8.0.16` while all six of its resolution entries say `8.0.8`.

Measured, by resetting the lockfile to main and installing each form:

| override form | resulting lockfile entries |
|---|---|
| `vite: ">=8.0.16"` | 6 × `vite@8.0.8` |
| `vite: 8.0.16` | 6 × `vite@8.0.16` |

## ws — two wrong turns, both recorded in the file

The audit reports two disjoint ranges on two paths (`@percolatorct/shared > ws` at `>=8.0.0 <8.21.0`, `@solana/web3.js > jayson > ws` at `>=7.0.0 <7.5.11`), so I first wrote a scoped pair to avoid dragging jayson from 7.x to 8.x.

Tracing the resolved graph showed that premise was false. The `<7.5.11` range only ever applied to *main's stale lockfile entry*; once the lockfile is refreshed, jayson reaches ws through `isomorphic-ws` and lands on **8.20.0 by itself**. So a single `>=8.21.0` floor moves it 8.20.0 → 8.21.3 — a patch bump inside the major it already uses — and there is no 7.x consumer left to protect. Confirmed afterwards: zero `ws@7` entries in the resolved graph.

I mention it because I twice drew a conclusion from a lockfile that a previous experiment had already mutated. `pnpm why` on a freshly-reset lockfile is the only reading that means anything.

## bigint-buffer stays suppressed, explicitly

GHSA-3gc7-fjrx-p6mg is genuinely unfixable: the advisory lists no patched version (`<0.0.0`) and bigint-buffer last published 1.1.5 on **2019-10-17**. It is suppressed via `pnpm.auditConfig.ignoreGhsas` rather than by lowering `--audit-level` or leaving the gate off, so **every other high still blocks**. The reasoning lives in a comment above the audit step because `package.json` cannot carry comments. #372 stays open to track it.

(#372 was filed 2026-06-27 saying a patched release might land; I re-checked the registry today — still 1.1.5, latest publish 2019. Its suggestion #2, a periodic audit check, is what this PR implements.)

## Verification

Run with **CI's own command**, not just a local install — an earlier local-only check passed while `--frozen-lockfile` would have failed, because plain `pnpm install` had refreshed versions the lockfile did not carry.

From a lockfile reset to main:

```
pnpm install --frozen-lockfile   exit 0
pnpm audit --audit-level=high    exit 0   (1 high — the ignored bigint-buffer)
pnpm build                       exit 0
pnpm test                        102 files, 1078 passed, 33 skipped
pnpm exec tsc --noEmit           exit 0
```

Resolved: `vite@8.0.16`, `ws@8.21.3`, `brace-expansion@5.0.9`, `nanoid@3.3.18`, `postcss@8.5.26`.

## Mutation-tested

A gate that cannot fail is not a gate. Each round resets the lockfile to main and re-installs, with a control on both sides:

| round | audit exit |
|---|---|
| control (shipped) | **0** |
| vite pin back to `8.0.8` | **1** |
| `ws` floor removed | **1** |
| ignore list emptied | **1** |
| control (shipped) | **0** |

Every element is load-bearing, and the gate blocks a real high rather than passing vacuously.

## Note on scope

`postcss` and `nanoid` are build-time only — the Dockerfile installs production dependencies only, and the `docker` job added in #391 asserts that. They are floored to keep the gate green, not because they are reachable in production. `ws` and `brace-expansion` are runtime-reachable.

Needs Security review in particular: this changes what ships in the runtime image.

## Summary by CodeRabbit

* **Security**
  * Updated vulnerable dependencies to secure minimum versions.
  * Added automated high-severity dependency auditing to the build process.
  * Documented an explicitly accepted security advisory exception.
  * The audit continues to block builds for other high-severity vulnerabilities.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,28 @@ jobs:
           node-version: 22
           cache: pnpm
       - run: pnpm install --frozen-lockfile
+      # #372: the keeper had no dependency-audit step at all. Every CVE fixed in
+      # this repo so far — #378's @opentelemetry/core, #407's ws — was found by
+      # hand, which means a regression, or a newly published advisory against a
+      # dependency we already ship, could sit here indefinitely. This is the
+      # service that holds a hot signing wallet, and it was the last of the four
+      # repos without a gate.
+      #
+      # One HIGH advisory is suppressed via pnpm.auditConfig.ignoreGhsas in
+      # package.json (which cannot carry comments), so the reasoning lives here:
+      #
+      #   GHSA-3gc7-fjrx-p6mg — buffer overflow in bigint-buffer's toBigIntLE().
+      #   Reached only as @percolatorct/sdk > @solana/spl-token >
+      #   @solana/buffer-layout-utils > bigint-buffer. The advisory lists NO
+      #   patched version ("<0.0.0"), and bigint-buffer's most recent release is
+      #   1.1.5 from 2019-10-17 — the package is unmaintained, so there is
+      #   nothing to override to and nothing in our control drops it.
+      #
+      # Suppressing it explicitly, rather than lowering --audit-level or leaving
+      # the gate off, keeps the exposure recorded and keeps every OTHER high
+      # advisory blocking. Revisit when @solana/spl-token drops
+      # @solana/buffer-layout-utils; #372 tracks it.
+      - run: pnpm audit --audit-level=high
       - run: pnpm build
       - run: pnpm test
 
--- a/package.json
+++ b/package.json
@@ -28,5 +28,12 @@
     "tsx": "^4.19.2",
     "typescript": "^5.7.3",
     "vitest": "^4.0.18"
+  },
+  "pnpm": {
+    "auditConfig": {
+      "ignoreGhsas": [
+        "GHSA-3gc7-fjrx-p6mg"
+      ]
+    }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,9 +5,10 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  vite: 8.0.8
+  vite: 8.0.16
   '@opentelemetry/core': '>=2.8.0'
   ws: ^8.21.0
+  brace-expansion: '>=5.0.9'
 
 importers:
 
@@ -55,7 +56,7 @@ importers:
         version: 5.9.3
       vitest:
         specifier: ^4.0.18
-        version: 4.1.2(@opentelemetry/api@1.9.1)(@types/node@25.5.0)(vite@8.0.8(@types/node@25.5.0)(esbuild@0.27.4)(tsx@4.21.0))
+        version: 4.1.2(@opentelemetry/api@1.9.1)(@types/node@25.5.0)(vite@8.0.16(@types/node@25.5.0)(esbuild@0.27.4)(tsx@4.21.0))
 
 packages:
 
@@ -63,11 +64,11 @@ packages:
     resolution: {integrity: sha512-JiDShH45zKHWyGe4ZNVRrCjBz8Nh9TMmZG1kh4QTK8hCBTWBi8Da+i7s1fJw7/lYpM4ccepSNfqzZ/QvABBi5g==}
     engines: {node: '>=6.9.0'}
 
-  '@emnapi/core@1.9.2':
-    resolution: {integrity: sha512-UC+ZhH3XtczQYfOlu3lNEkdW/p4dsJ1r/bP7H8+rhao3TTTMO1ATq/4DdIi23XuGoFY+Cz0JmCbdVl0hz9jZcA==}
+  '@emnapi/core@1.10.0':
+    resolution: {integrity: sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==}
 
-  '@emnapi/runtime@1.9.2':
-    resolution: {integrity: sha512-3U4+MIWHImeyu1wnmVygh5WlgfYDtyf0k8AbLhMFxOipihf6nrWC4syIm/SwEeec0mNSafiiNnMJwbza/Is6Lw==}
+  '@emnapi/runtime@1.10.0':
+    resolution: {integrity: sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==}
 
   '@emnapi/wasi-threads@1.2.1':
     resolution: {integrity: sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==}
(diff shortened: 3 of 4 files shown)


dcccrypto:
**Correcting my own risk framing in this PR, prompted by #407.**

The `pnpm-workspace.yaml` comment here says the `ws` floor moves jayson "8.20.0 -> 8.21.3, a patch bump within the major it already uses, and no 7.x consumer is left to protect." That is measured against an intermediate refreshed lockfile that **never ships**, and it understates what this PR actually does.

Measured against `main`'s committed lockfile — what is running today:

```
main   ws@7.5.10   ws@8.20.0
#406   ws@8.21.3   (7.5.10 gone)
```

So this PR **does** move jayson's path from ws 7 to ws 8, a major version, exactly as #407 says. My reasoning was right about the resolution mechanics (a refresh alone lands jayson on 8.20.0) but wrong about the baseline: the refresh is part of *this PR*, so a reviewer comparing against what is deployed sees 7 -> 8. #407's residual-risk note is the more honest framing of the same change, and I'd rather say so here than leave a comment in the tree that reads more reassuring than the facts support.

**That said, I went and bounded the risk, and it is smaller than either of us stated.**

The concern is that a jayson ws major bump breaks the `@solana/web3.js` WebSocket RPC path, which no test exercises. I checked whether this service drives that path at all:

- **No WebSocket subscription APIs anywhere in `src`** — zero hits for `onAccountChange`, `onProgramAccountChange`, `onLogs`, `onSlotChange`, `removeAccountChangeListener`, `wsEndpoint`.
- **No `confirmTransaction` and no `onSignature`**, in `src` or in `@percolatorct/shared`'s `dist`. That matters most: `confirmTransaction` is the usual way a service opens a signature subscription without meaning to. Confirmation here is polled instead — `pollSignatureStatus` and `getSignatureStatuses` (`rpc-pool.ts:311`, `keeper-send.ts:177`).
- **LaserStream is gRPC, not ws** (`account-loader.ts:34`, "Helius LaserStream gRPC endpoint"), so it is not on this path either.

web3.js opens its RPC WebSocket lazily, on first subscription. With no subscription call sites and no `confirmTransaction`, the jayson -> isomorphic-ws -> ws path is a dependency this service carries but never drives.

So the honest statement is: **the graph change is a major bump; the exercised runtime surface for it appears to be zero.** Not "proven safe by tests" — there are none — but bounded by call-site absence, which is checkable and which I checked.

**Overlap with #407, needs a call from whoever merges.** #407 makes the same `ws` change and nothing else; this PR includes it plus the vite unpin, the `bigint-buffer` suppression and the audit gate, so it is a superset and the two will conflict on `pnpm-lock.yaml` whichever lands second.

I am not closing anyone's PR. My suggestion is this one lands and #407 / #371 close as subsumed, with the CVE credit belonging to #371 which found them — but if you would rather take the narrow #407 first, say so and I will rebase this onto it promptly rather than leave it stale.

dcccrypto:
**Rebased onto `a0971ab` now that #407 has landed** — as offered. Re-measured from the new base rather than carrying the previous version's conclusions across, which turned out to matter: the right change on this base is smaller than the one I had.

**What changed in the PR**

- **`ws` floor dropped.** #407 fixed it on main (`ws: ^8.21.0`, resolving to 8.21.3), so this no longer needs to touch it. Credit for that one belongs to #371/#407.
- **`nanoid` and `postcss` floors dropped.** Both were flagged (2 highs each, build-time only via `vitest > vite`), but on this base a plain `pnpm install` already resolves them to 3.3.18 and 8.5.26 within the ranges their parents ask for. Pinning them would record a requirement nothing is enforcing.
- **`brace-expansion` floor kept** — it is the one that does *not* move on its own. It stays at 5.0.5 and needs `>=5.0.9`. Measured, not assumed: with the vite fix and a refresh but no floors, the audit still reports 3 highs, all brace-expansion.

**Unchanged, and still the point of the PR:** main has no audit step and `vite: 8.0.8` is still there — an exact pin added before GHSA-fx2h-pf6j-xcff published, holding vite *below* the fix. Turning the gate on against current main reports **9 high** advisories.

**Verification, from a lockfile reset to main, using CI's own command**

```
pnpm install --frozen-lockfile   exit 0
pnpm audit --audit-level=high    exit 0   (1 high — the ignored bigint-buffer)
pnpm build                       exit 0
pnpm test                        105 files, 1121 passed, 33 skipped
pnpm exec tsc --noEmit           exit 0
```

**Mutation battery re-run on this base** — the previous run proved the old config, not this one, so it does not transfer:

| round | audit exit |
|---|---|
| control (shipped) | **0** |
| vite pin back to `8.0.8` | **1** |
| `brace-expansion` floor removed | **1** |
| ignore list emptied | **1** |
| control (shipped) | **0** |

Each round resets the lockfile to main and re-installs with a checked exit code, with a control either side.

**On the correction I posted earlier**: it stands, and it now applies to what merged rather than to this PR. #407 is what moved jayson's path from `ws@7.5.10` to `8.21.3` — a major version — and my earlier "patch bump" framing of that change was measured against a baseline that never shipped. The bounding evidence still holds and is worth keeping visible for whoever looks at the ws change later: this service has no web3.js WebSocket subscription call sites at all — no `onAccountChange`, `onProgramAccountChange`, `onLogs`, `onSlotChange`, `wsEndpoint`, and no `confirmTransaction` or `onSignature` in `src` or in `@percolatorct/shared`'s `dist`. Confirmation is polled (`pollSignatureStatus`, `getSignatureStatuses`). LaserStream is gRPC. web3.js opens its RPC socket lazily on first subscription, so it is never opened here.

One small thing for a follow-up, not worth a PR on its own: the `ws` comment that landed in `pnpm-workspace.yaml` cites CVE-2024-37890 / GHSA-3h5v-q93c-6h6q (`ws < 8.17.1`, one object per HTTP header). The advisory the audit actually reports for those paths is **GHSA-96hv-2xvq-fx4p** (memory exhaustion from tiny fragments, `>=8.0.0 <8.21.0` and `>=7.0.0 <7.5.11`) — which is what makes 8.20.0 vulnerable at all and what sets the 8.21.0 floor. The floor chosen is correct; only the advisory named is wrong. Happy to fix that comment in this PR if you want it folded in.

PR #407, opened 2026-08-09
fix(deps): override ws to ^8.21.0 (re-cut of #371 against current main)

dcccrypto:
Re-cut of #371. That PR's lockfile predates the SDK re-pin to `886aa6b` and #386's `@opentelemetry/core` override, so it conflicts on both `pnpm-lock.yaml` and `pnpm-workspace.yaml` and cannot be rebased mechanically.

`ws < 8.17.1` allocates one object per HTTP header, so a request with many headers exhausts memory (CVE-2024-37890). The keeper reaches `ws` two ways:

- `@supabase/realtime-js` and `helius-laserstream` → `ws@8.20.0`
- `@solana/web3.js` → `jayson` → `isomorphic-ws` → `ws@7.5.10`

The keeper has no direct `ws` dependency, so an override is the only lever. After it the tree collapses to a single `ws@8.21.3`; the only remaining `ws@7` entries are `@types/ws@7.4.7`, which is types-only.

**Residual risk for a reviewer:** this bumps `jayson`'s `ws` from 7 to 8, a major version. Its runtime use through web3.js is the basic client surface (connect/send/close/on), the build is clean and the full suite passes — but nothing in the suite exercises the web3.js WebSocket RPC path, so that route is untested here rather than proven.

Verified: `pnpm install --frozen-lockfile` clean, `tsc` clean, **102 files / 1078 passed / 33 skipped**.

Closes #371.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,6 +7,7 @@ settings:
 overrides:
   vite: 8.0.8
   '@opentelemetry/core': '>=2.8.0'
+  ws: ^8.21.0
 
 importers:
 
@@ -1067,7 +1068,7 @@ packages:
   isomorphic-ws@4.0.1:
     resolution: {integrity: sha512-BhBvN2MBpWTaSHdWRb/bwdZJ1WaehQ2L1KngkCkfLUGF0mAWAT1sQUQacEmQ0jXkFw/czDXPNQSL5u2/Krsz1w==}
     peerDependencies:
-      ws: '*'
+      ws: ^8.21.0
 
   jayson@4.3.0:
     resolution: {integrity: sha512-AauzHcUcqs8OBnCHOkJY280VaTiCm57AbuO7lqzcw7JapGj50BisE3xhksye4zlTSR1+1tAz67wLTl8tEH1obQ==}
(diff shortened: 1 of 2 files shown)


dcccrypto:
Flagging an overlap, and adding evidence that bounds the residual risk you called out.

**Overlap.** #406 makes the same `ws` override, plus the vite unpin, the `bigint-buffer` suppression and a `pnpm audit --audit-level=high` gate in CI. Both PRs touch `pnpm-workspace.yaml` and `pnpm-lock.yaml`, so whichever lands second will conflict on the lockfile. #406 is a superset of this one — worth a merge-order decision rather than discovering it as a conflict.

I am not closing this. If the narrow change is preferred first, I will rebase #406 onto it.

**On the residual risk you flagged** — "this bumps jayson's `ws` from 7 to 8 ... nothing in the suite exercises the web3.js WebSocket RPC path, so that route is untested here rather than proven". That framing is right, and it is more honest than the one I had written into #406, which described the same change as a patch bump by comparing against a refreshed lockfile that never ships. Against `main`'s committed lockfile it is 7.5.10 -> 8.21.3, a major, exactly as you say. I have corrected my own PR.

I then went looking for how much runtime surface that actually has, and it appears to be none:

- **No WebSocket subscription APIs anywhere in `src`** — zero hits for `onAccountChange`, `onProgramAccountChange`, `onLogs`, `onSlotChange`, `removeAccountChangeListener`, `wsEndpoint`.
- **No `confirmTransaction`, no `onSignature`** — in `src` or in `@percolatorct/shared`'s `dist`. This is the one I most expected to find, since `confirmTransaction` opens a signature subscription without the caller thinking about WebSockets. Confirmation is polled instead: `pollSignatureStatus`, `getSignatureStatuses` (`rpc-pool.ts:311`, `keeper-send.ts:177`).
- **LaserStream is gRPC, not ws** (`account-loader.ts:34`), so it is not on the jayson path either.

web3.js opens its RPC WebSocket lazily on first subscription. With no subscription call sites and no `confirmTransaction`, that socket is never opened by this service — the jayson -> isomorphic-ws -> ws dependency is carried but never driven.

That does not make it "proven safe" — there are still no tests on that path, and your wording of that is exactly right. It bounds it differently: not "the tests pass so it is probably fine", but "the call sites that would exercise it do not exist", which is checkable and which I checked. Worth adding to whichever of these two PRs ends up merging, since it is the reviewer's main open question either way.

One small correction to the description here, unrelated to the risk: the body cites CVE-2024-37890 (`ws < 8.17.1`, one object per HTTP header). The advisory the audit currently reports for these paths is GHSA-96hv-2xvq-fx4p — memory exhaustion from tiny fragments — covering `>=8.0.0 <8.21.0` and `>=7.0.0 <7.5.11`. That is what makes `8.20.0` and `7.5.10` both vulnerable and what sets the 8.21.0 floor; under the older CVE alone, 8.20.0 would already have been patched. Same fix, and the floor you chose is correct — just a different advisory than the one named.

PR #411, opened 2026-08-09
fix(env-guards): stop echoing CRANK_KEYPAIR secret bytes in the boot error (closes #410)

### LINKED ISSUE #410: keeper: malformed CRANK_KEYPAIR echoes secret-key bytes into the boot error
## Summary

The boot-time `CRANK_KEYPAIR` format check re-throws with the underlying parse error appended. For the JSON-array keypair form, `loadKeypair` calls `JSON.parse`, whose `SyntaxError` message quotes an excerpt of the input — i.e. **raw secret-key bytes** — so a malformed keypair leaks part of the private key into the boot error, which lands in stderr, Railway logs, and Sentry.

**Severity: Medium** — private-key material disclosure into logs/monitoring. Triggered by exactly the case the guard exists for: a truncated or typo'd keypair.

## Root cause

`src/env-guards.ts`:

```ts
try {
  loadKeypair(crankKp);
} catch (err) {
  throw new Error(
    `CRANK_KEYPAIR is not a valid keypair (expected a 64-byte JSON array or ` +
      `base58-encoded secret key): ${err instanceof Error ? err.message : String(err)}`,
  );
}
```

For `CRANK_KEYPAIR="[12,84,201,7,55,163,9,241,…"` (truncated/typo'd), Node's `JSON.parse` throws e.g. `Unexpected token …, "[12,84,201,7,55,163,9…" is not valid JSON` — the quoted excerpt is secret-key bytes, which the re-thrown message then carries into logs and Sentry.

## Fix

Drop the appended parse error; throw a static, actionable message. The presence of secret bytes in the parse error is exactly why it must not be logged.

## Proof of Concept

`tests/env-guards.test.ts` (M1 block):

```ts
it("M1: does not echo CRANK_KEYPAIR secret bytes into the boot error", () => {
  // A malformed JSON secret-key array. JSON.parse throws a SyntaxError that
  // quotes an excerpt of the input — real secret-key bytes — pre-fix.
  const env = { CRANK_KEYPAIR: "[201,163,88,7,x]" } as NodeJS.ProcessEnv;
  let caught: Error | undefined;
  try {
    validateKeeperEnvGuards(env);
  } catch (e) {
    caught = e as Error;
  }
  expect(caught).toBeInstanceOf(Error);
  expect(caught!.message).toMatch(/CRANK_KEYPAIR is not a valid keypair/);
  // The raw key bytes must not appear anywhere in the message.
  expect(caught!.message).not.toContain("201,163,88");
  expect(caught!.message).not.toContain("[201");
});
```

Fails before the fix (the message contains `"[201,163,88,7,x]"` from the `SyntaxError`); passes after. The existing M1 tests pin only the static `/CRANK_KEYPAIR is not a valid keypair/` prefix, which the fix preserves.

0x-SquidSol:
Closes #410.

## Problem

The boot-time `CRANK_KEYPAIR` format check re-threw with the underlying parse error appended. For the JSON-array keypair form, `loadKeypair` calls `JSON.parse`, whose `SyntaxError` quotes an excerpt of the input — **raw secret-key bytes** — so a truncated or typo'd keypair leaked part of the private key into the boot error, and thus into stderr, Railway logs, and Sentry.

## Fix

Drop the appended parse error and throw a static, actionable message. The static `CRANK_KEYPAIR is not a valid keypair …` prefix that the existing M1 tests pin is preserved, so no pinned behavior changes.

## Testing

- New PoC asserts a malformed keypair (`[201,163,88,7,x]`) throws an error whose message does **not** contain the byte excerpt (fails before: the message carried `"[201,163,88,7,x]"` from the `SyntaxError`).
- `tests/env-guards.test.ts` green (77); `tsc --noEmit` clean.

### DIFF
--- a/src/env-guards.ts
+++ b/src/env-guards.ts
@@ -81,10 +81,15 @@ export function validateKeeperEnvGuards(env: NodeJS.ProcessEnv = process.env): v
   if (crankKp) {
     try {
       loadKeypair(crankKp);
-    } catch (err) {
+    } catch {
+      // Do NOT append the underlying parse error. For the JSON-array form,
+      // JSON.parse quotes an excerpt of the input — i.e. raw secret-key bytes —
+      // into its SyntaxError message, which would then land in stderr / Railway
+      // logs / Sentry. A truncated or typo'd keypair is exactly this case.
       throw new Error(
-        `CRANK_KEYPAIR is not a valid keypair (expected a 64-byte JSON array or ` +
-          `base58-encoded secret key): ${err instanceof Error ? err.message : String(err)}`,
+        "CRANK_KEYPAIR is not a valid keypair (expected a 64-byte JSON array " +
+          "from `solana-keygen new`, or a base58-encoded secret key). The " +
+          "underlying parse error is omitted because it can contain secret-key bytes.",
       );
     }
   }
--- a/tests/env-guards.test.ts
+++ b/tests/env-guards.test.ts
@@ -436,6 +436,23 @@ describe("validateKeeperEnvGuards", () => {
         /CRANK_KEYPAIR is not a valid keypair/,
       );
     });
+    it("M1: does not echo CRANK_KEYPAIR secret bytes into the boot error", () => {
+      // A malformed JSON secret-key array. JSON.parse throws a SyntaxError that
+      // quotes an excerpt of the input — real secret-key bytes — pre-fix. The
+      // boot error must not carry them into stderr / Railway logs / Sentry.
+      const env = { CRANK_KEYPAIR: "[201,163,88,7,x]" } as NodeJS.ProcessEnv;
+      let caught: Error | undefined;
+      try {
+        validateKeeperEnvGuards(env);
+      } catch (e) {
+        caught = e as Error;
+      }
+      expect(caught).toBeInstanceOf(Error);
+      expect(caught!.message).toMatch(/CRANK_KEYPAIR is not a valid keypair/);
+      // The raw key bytes must not appear anywhere in the message.
+      expect(caught!.message).not.toContain("201,163,88");
+      expect(caught!.message).not.toContain("[201");
+    });
 
     it("M1: accepts a well-formed 64-byte JSON array keypair", () => {
       const env = { CRANK_KEYPAIR: VALID_KEYPAIR_JSON } as NodeJS.ProcessEnv;


dcccrypto:
Reviewed against `a0971ab`. The leak is real, the fix closes it, and the test binds. Evidence rather than assertion, since this one is a private-key disclosure.

**Reproduced the leak.** Reverting `src/env-guards.ts` to main's version and running the new test produces the message verbatim:

```
CRANK_KEYPAIR is not a valid keypair (expected a 64-byte JSON array or
base58-encoded secret key): Unexpected token 'x', "[201,163,88,7,x]" is not valid JSON
```

The input bytes are quoted straight into the boot error, exactly as #410 describes. This is not a theoretical path — it fires on precisely the case the guard exists to catch (truncated or typo'd keypair), and lands in stderr, Railway logs and Sentry.

**The test binds.** Mutation with a control either side:

| | result |
|---|---|
| fix in place | 77 passed |
| `env-guards.ts` reverted to main | **1 failed** — `expected '…' not to contain '201,163,88'` |
| fix restored | 77 passed |

Both assertions earn their place: `not.toContain("201,163,88")` catches the quoted excerpt and `not.toContain("[201")` catches a leading-bracket variant. Asserting on the *absence* of a substring is usually a weak test, but here the mutation shows it fails for the right reason, which is what makes it meaningful.

**Full suite on the merge:** 105 files, 1122 passed, 33 skipped; `tsc --noEmit` exit 0.

**Variant hunt — I went looking for siblings of this bug class, and there is one worth naming.**

`loadKeypair` has two call sites, and this PR hardens one of them:

- `src/env-guards.ts:83` — wrapped, now redacted. ✅
- `src/lib/keypair-singleton.ts:33` — `loadKeypair(process.env.CRANK_KEYPAIR!)`, **unwrapped**. If it throws, that same byte-quoting `SyntaxError` propagates uncaught.

It is not reachable today, and I checked rather than assumed: `index.ts` is the only entrypoint (`start: node dist/index.js`), and it orders presence check (`:68`) → `validateKeeperEnvGuards()` (`:71`) → `getKeeperKeypair()` (`:81`) → services (`:111-112`). So the singleton only ever parses a value the guard already accepted.

But that safety is a property of the ordering in `index.ts`, not of the module — and the call site is one day old (#355). Worth one line of defence or at minimum a comment on `keypair-singleton.ts` recording that it relies on the guard having run first, so a future second entrypoint does not quietly reintroduce this. Not blocking, and arguably out of scope here.

**One thing I suspected and could not substantiate**, recorded so nobody re-treads it: the guard validates `env.CRANK_KEYPAIR?.trim()` while the singleton parses the **untrimmed** `process.env.CRANK_KEYPAIR!` — two different strings, so in principle the guard could bless one value while a different one is parsed. I probed it with a generated keypair in clean, space-padded and trailing-newline forms; `loadKeypair` accepted all three and the guard accepted all three. No divergence. The mismatch is real in the code but I could not turn it into a defect.

**On the message itself:** dropping the underlying error is the right call, and the replacement stays actionable by naming `solana-keygen new` and both accepted formats — an operator hitting this still knows what to do, which is the usual cost of redacting an error and it has been paid here.

Not approving formally — this touches the signing key and should carry Security sign-off. Nothing from my side blocks it.

PR #413, opened 2026-08-09
fix(health): mask the RPC api-key in /health failureReason (closes #412)

### LINKED ISSUE #412: keeper: /health failureReason can leak the RPC URL + api-key
## Summary

When the boot-time primary-RPC connectivity probe fails, the error message — a node-fetch error that quotes the full RPC URL, including the `?api-key=…` query param — is stored as `startupTracker.failureReason` and served verbatim in the `/health` 503 body (and sent to Sentry). Any prober of `/health` during a boot RPC failure (or on each crash-loop restart) receives the RPC api-key.

**Severity: Medium** — RPC api-key disclosure. The exposure is worse under the Railway remote-bind posture where `/health` is reachable without a secret.

## Root cause

`src/index.ts` — the RPC probe catch:

```ts
} catch (err) {
  const msg = err instanceof Error ? err.message : String(err);   // "request to https://…?api-key=XXX failed …"
  logger.error("Primary RPC unreachable at startup — check SOLANA_RPC_URL", { error: msg });
  throw new Error(`Primary RPC connectivity check failed: ${msg}`);
}
```

The thrown error's message flows to `startupTracker.markFailed(err.message)` and is served as `failureReason` at the `/health` 503 branch, and to Sentry via `captureAndExit(..., err, ...)`.

## Fix

Reuse the existing `maskApiKeys` helper from `@percolatorct/shared` (already in the installed dist):
- In `StartupTracker.markFailed` — so anything served as `failureReason` on `/health` is masked, for this and any future failure path.
- At the RPC-probe source in `index.ts` — so the log line and the thrown error (→ Sentry) are masked too.

No new code and no repin — `maskApiKeys` replaces `api[_-]?key=…` with `api-key=***`.

## Proof of Concept

`tests/lib/startup-tracker.test.ts`:

```ts
it("masks an embedded api-key in the failure reason (served on /health)", () => {
  const t = new StartupTracker();
  // A boot RPC failure surfaces as a node-fetch error quoting the full URL.
  t.markFailed(
    "Primary RPC connectivity check failed: request to " +
      "https://mainnet.helius-rpc.com/?api-key=SECRET123 failed, reason: ECONNREFUSED",
  );
  // failureReason is returned in the /health 503 body — it must not leak the key.
  expect(t.failureReason).not.toContain("SECRET123");
  expect(t.failureReason).toContain("api-key=***");
});
```

Fails before the fix (`failureReason` contains `SECRET123`); passes after.

0x-SquidSol:
Closes #412.

## Problem

When the boot-time primary-RPC connectivity probe fails, the node-fetch error quotes the full RPC URL including `?api-key=…`. That message was stored as `startupTracker.failureReason` and served verbatim in the `/health` 503 body (and sent to Sentry), so any prober during a boot RPC failure — or on each crash-loop restart — received the RPC api-key. The exposure is worse under the Railway remote-bind posture where `/health` is reachable without a secret.

## Fix

Reuse `maskApiKeys` from `@percolatorct/shared` (already in the installed dist — replaces `api[_-]?key=…` with `api-key=***`):
- `StartupTracker.markFailed` masks the reason before storing, so anything served as `failureReason` on `/health` is clean — for this and any future failure path.
- The RPC-probe catch in `index.ts` masks at the source, so the log line and the thrown error (→ Sentry via `captureAndExit`) are masked too.

No new code, no repin.

## Testing

- New PoC asserts `markFailed(<error with ?api-key=SECRET123>)` yields a `failureReason` that masks the key (`api-key=***`, no `SECRET123`) — fails before the fix.
- `tests/lib/startup-tracker.test.ts` green (7); `tsc --noEmit` clean.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import "dotenv/config";
 import http from "node:http";
 import { timingSafeEqual } from "node:crypto";
-import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, getConnection, loadKeypair } from "@percolatorct/shared";
+import { config, createLogger, initSentry, captureException, sendInfoAlert, sendCriticalAlert, sendWarningAlert, getConnection, loadKeypair, maskApiKeys } from "@percolatorct/shared";
 import { monitors } from "./lib/service-monitors.js";
 import { OracleService } from "./services/oracle.js";
 import { CrankService } from "./services/crank.js";
@@ -765,7 +765,10 @@ async function start() {
       });
     }
   } catch (err) {
-    const msg = err instanceof Error ? err.message : String(err);
+    // Mask any embedded api-key: a node-fetch failure quotes the full RPC URL
+    // ("request to https://…?api-key=XXX failed"), and this msg flows to the log,
+    // the thrown error (→ Sentry via captureAndExit), and /health's failureReason.
+    const msg = maskApiKeys(err instanceof Error ? err.message : String(err));
     logger.error("Primary RPC unreachable at startup — check SOLANA_RPC_URL", { error: msg });
     throw new Error(`Primary RPC connectivity check failed: ${msg}`);
   }
--- a/src/lib/startup-tracker.ts
+++ b/src/lib/startup-tracker.ts
@@ -1,3 +1,5 @@
+import { maskApiKeys } from "@percolatorct/shared";
+
 /**
  * Tracks whether the asynchronous keeper boot (RPC connectivity probe,
  * market discovery, service start) has completed.
@@ -19,7 +21,11 @@ export class StartupTracker {
 
   markFailed(reason: string): void {
     this._state = "failed";
-    this._failureReason = reason;
+    // failureReason is served on /health (503 body), so mask any embedded
+    // api-key before storing it. A boot RPC failure surfaces as a node-fetch
+    // error quoting the full URL ("request to https://…?api-key=XXX failed"),
+    // which would otherwise leak the RPC key to any prober and to Sentry.
+    this._failureReason = maskApiKeys(reason);
   }
 
   get state(): StartupState {
--- a/tests/lib/startup-tracker.test.ts
+++ b/tests/lib/startup-tracker.test.ts
@@ -50,4 +50,16 @@ describe("StartupTracker", () => {
     t.markReady();
     expect(t.state).toBe("ready");
   });
+
+  it("masks an embedded api-key in the failure reason (served on /health)", () => {
+    const t = new StartupTracker();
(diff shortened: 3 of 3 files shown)


dcccrypto:
Heads-up while this is adjacent to #411, which I just reviewed: this PR is currently **DIRTY** and `build-and-test` has not run on it — only CodeRabbit reported. Main moved a fair way in the last couple of hours (#409, #411, #415 and the earlier batch), so it likely needs a rebase before it can be assessed at all.

Worth saying that its sibling #411 checks out: I reproduced the `CRANK_KEYPAIR` leak verbatim, the fix closes it and its test binds under mutation. Same bug class as this one — a secret reaching a log/monitoring sink through an error string — so this is worth unblocking rather than leaving to go stale.

One suggestion from reviewing #411, if it helps here: an absence assertion (`not.toContain(apiKey)`) only means something if you have shown it fails when the redaction is removed. Mutating the fix out and watching that specific assertion fail is what turns it from a test that always passes into one that binds.

dcccrypto:
Rebased onto current `main` and pushed the resolution — this now merges cleanly.

The conflict was a one-line import collision, not a design clash: this PR adds `maskApiKeys` to the `@percolatorct/shared` import, and #355 removed `loadKeypair` from that same line (it moved to `getKeeperKeypair`). Kept both changes.

Verified on the merge: `tsc` clean, full suite **105 files / 1122 passed / 33 skipped**. `maskApiKeys` does resolve from the package root — shared's `dist/index.js` has `export * from "./sanitize.js"`.

Worth noting this is complementary to #359 rather than overlapping: #359 reduces the `/health` payload for unauthenticated remote callers, but `failureReason` is served from the startup-503 branch, which sits *above* that exposure gate — so #359 did not cover this path.

dcccrypto:
**Correction to my previous comment.** I said I had pushed the resolution to this branch — that push was rejected (`permission denied`) despite `maintainerCanModify: true`, so this branch is unchanged and still conflicting. Apologies for the noise.

The rebase is up as **#424** instead, with no functional change from this PR — same fix, same tests, conflict resolved. Verified there: `tsc` clean, full suite 105 files / 1122 passed / 33 skipped.

Close this in favour of #424, or re-push the rebase here yourself if you'd rather keep authorship on this branch — the resolution is just keeping `maskApiKeys` on the shared import line while dropping `loadKeypair` (which #355 moved to `getKeeperKeypair`).

dcccrypto:
Closing in favour of #424, which is this exact change rebased onto current `main` (merged as 3a10d8a). Authorship note: I could not push the rebase to this branch — the push was rejected with `permission denied` despite `maintainerCanModify: true` — so a new PR from the origin repo was the only route. The fix and its test are yours, unchanged.

Issue #426, opened 2026-08-10
[Security Hardening] Priority-Fee Fallback and Silent Stream-Stall Cache Freshness


Bayyan16:
### Verdict: no High/Critical vulnerability on HEAD · two Low-severity hardening items · reproducible verification PoC included

---

| Field | Value |
|---|---|
| **Commit audited** | `052b0c53c8cb7d1738160960d30e3dc4cc33598b` (`main`, HEAD) |
| **Scope** | 100% of `src/` — all 46 modules (7 services + 39 lib/config modules) |
| **Method** | Local clone → reproducible sandbox (`pnpm install`, TypeScript build, `vitest`) → full manual source review → property verification against the real compiled code |
| **Result** | **No novel High/Critical found.** 2 Low-severity / defense-in-depth observations. 2 documented residual tradeoffs (not keeper-code bugs). |

---

## 1. Executive summary

This is a whole-codebase security review of the Percolator keeper at HEAD. The keeper is **off-chain** infrastructure: it monitors accounts and submits liquidation, oracle-price-push, and funding/settlement crank transactions. It cannot move user funds directly — the on-chain program is the authority — so the review is weighted toward the harms a compromised or buggy keeper *can* cause: missed liquidations (protocol bad debt), a manipulated price pushed on-chain, keeper-wallet fee drain, HA split-brain, and endpoint/secret exposure.

**The headline finding is a negative one, and it is deliberate.** The keeper has already been through many rounds of PoC-driven hardening (dozens of closed issues in the `#297`–`#425` range plus labelled batches `C2`–`C4`, `H1`–`H8`, `M-x`, `#329`–`#342`). Every high-value attack surface examined is not only hardened but hardened *with committed PoC tests*, which this audit re-ran and confirmed still pass. Reporting a fabricated or inflated "High" would violate the non-inflation bar this methodology is built on: trace-to-live-code, PoC-verify, confirm non-duplicate, rate severity accurately, and be willing to conclude "nothing High/Critical" when that is the truth.

What the audit *does* surface is two **Low-severity, defense-in-depth** robustness items (§4.1, §4.2) and two **documented residual tradeoffs** that belong to the broader architecture rather than keeper code (§4.3, §4.4). None is a vulnerability; all four are stated at their true severity.

[...]

PR #427, opened 2026-08-10
fix(crank): treat a silently stalled stream as stale cache, not fresh (#426 5.2)

dcccrypto:
Implements §5.2 of #426. Leaving §5.1 (priority-fee ceiling / tier-aware fallback) alone for now — see the note at the bottom.

## The gap

The account cache expires entries by **slot** age, measured against the stream's own `lastSlot`. That is a sound freshness signal only while the stream is delivering.

If the gRPC connection stays up but stops producing — a silent stall — `lastSlot` freezes. The slots stamped on cached entries freeze with it, so `currentSlot - entry.slot` stops growing and the TTL **can never expire**. `getOwnerVerified()` keeps returning frozen bytes as fresh for as long as the stall lasts.

`onStreamError -> cache.invalidateAll()` doesn't help: it fires on an error event, and a silent stall raises none. The gap is exactly the case where nothing goes wrong loudly.

Verified against `052b0c5` before writing anything — `crank.ts:1016-1017` takes `currentSlot` from `getStats().lastSlot`, and `invalidateAll()` appears only inside `onStreamError` (`account-loader.ts:379/388`).

## The fix

`AccountLoader` records `lastSlotAdvanceAt` — wall-clock time its slot last moved **forward** — and exposes it on `LoaderStats`. The crank discover fast path skips cache reads entirely while that stamp is older than `KEEPER_STREAM_STALL_MS` (default **30s**: ~75 slots, far outside normal jitter, comfortably above the 32-slot/~13s TTL it guards, far below the 30-minute full re-discover).

A stalled stream then behaves exactly like a cache miss, which the loop already handles — each market keeps the state it had until the next full re-discover, which reads fresh RPC. No new failure mode, just no false freshness.

The floor is wall-clock rather than slot-based deliberately: slot time is the quantity that has stopped being trustworthy, so it can't measure its own staleness.

## One thing worth a reviewer's eye

**`lastSlot` is still assigned unconditionally, and there's a test pinning that.**

My first draft clamped it to forward-only movement. That reads as harmless tidying and is not: the cache's reorg invalidation (**A.2** — `currentSlot < entry.slot` is a miss, `account-cache.ts:58-62`) *depends* on `lastSlot` being able to move **backward**. Clamping it would have silently disabled that check, trading this staleness bug for a worse one.

Only the *stamp* is gated on forward movement — a stream stuck re-announcing one slot must not look healthy. There is now a test that fails on the clamp, so the next person tempted by the same tidy-up gets told.

## Scope

Per the audit, and re-confirmed: only the crank **discover fast path** reads this cache. The 30-minute re-discover uses fresh RPC, the crank *submit* re-reads `getSlot("processed")`, and **the liquidation path never touches this cache** — so there was no wrong-or-missed-liquidation consequence, and there is none introduced here. This is defence-in-depth, rated Low in the audit, and I agree with that rating.

## Existing tests changed

Three A.1 fast-path tests now set `lastSlotAdvanceAt` in their mocked stats. That is the point rather than an inconvenience: a live stream now has to *say* it is live, where previously a frozen stream was indistinguishable from a healthy one at that call site.

## Mutation-tested

| mutation | result |
|---|---|
| control | all pass |
| stamp on every slot (not just forward) | **FAIL** `account-loader` |
| clamp `lastSlot` forward (breaks A.2 reorg) | **FAIL** `account-loader` |
| `isStreamStalled` always false | **FAIL** `stream-staleness` |
| loose env parse (`0`/negative disables the guard) | **FAIL** `stream-staleness` |
| crank reads cache despite the stall | **FAIL** `crank` |
| control | all pass |

The `account-loader` test needs **fake timers** to bind, and says so in a comment. Without them all three `pushSlot` calls land in the same millisecond, the stamps compare equal, and the assertions pass against the stamp-always mutation. I only found that by running the mutation — the test looked fine.

## Verification

```
pnpm test                       108 files, 1143 passed, 33 skipped
pnpm exec tsc --noEmit          exit 0
pnpm build                      exit 0
pnpm audit --audit-level=high   exit 0
```

## On §5.1

Not touched here. Its suggested hardening — "make the fallback tier-aware (liquidation / ADL should fall back higher than crank / oracle)" — means choosing fee values per lane, which is an operator economics call rather than a code detail, and the audit itself notes the budget gate already prevents overspend so the residual is availability-only. Happy to implement once someone says what the per-lane fallbacks should be.

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects stalled data streams when slot updates stop advancing.
  * Prevents indefinitely serving cached account data during silent stream interruptions.
  * Automatically falls back to full data refresh to obtain current information.
  * Supports configurable stall detection thresholds, with safe defaults for invalid settings.

### DIFF
--- a/src/lib/account-loader.ts
+++ b/src/lib/account-loader.ts
@@ -23,6 +23,21 @@ export type UnsubscribeFn = () => void;
 export interface LoaderStats {
   connected: boolean;
   lastSlot: number;
+  /**
+   * `Date.now()` when `lastSlot` last moved FORWARD, or 0 if it never has.
+   *
+   * #426 (5.2): consumers use `lastSlot` as the yardstick for cache freshness,
+   * which is only meaningful while the stream is actually delivering. A silent
+   * stall — gRPC connection up, no messages — freezes `lastSlot`, and because
+   * cached entries' slots freeze with it, a slot-age TTL can never expire. The
+   * cache would serve frozen bytes as fresh indefinitely. `onStreamError` does
+   * not help: it fires on an error event, and a silent stall raises none.
+   *
+   * Exposing the wall-clock stamp lets a consumer notice that slot-based
+   * freshness has stopped meaning anything, without this class having to guess
+   * what "too stale" is for any particular caller.
+   */
+  lastSlotAdvanceAt: number;
   eventsReceived: number;
   eventsDropped: number;
   reconnectCount: number;
@@ -193,6 +208,8 @@ export class AccountLoader {
   private running = false;
   private connected = false;
   private lastSlot = 0;
+  /** See LoaderStats.lastSlotAdvanceAt (#426). */
+  private lastSlotAdvanceAt = 0;
   private eventsReceived = 0;
   private eventsDropped = 0;
   private reconnectCount = 0;
@@ -264,6 +281,7 @@ export class AccountLoader {
     return {
       connected: this.connected,
       lastSlot: this.lastSlot,
+      lastSlotAdvanceAt: this.lastSlotAdvanceAt,
       eventsReceived: this.eventsReceived,
       eventsDropped: this.eventsDropped,
       reconnectCount: this.reconnectCount,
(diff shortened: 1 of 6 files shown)


dcccrypto:
Cross-check against keeper #428, which landed in the queue after this and also edits `src/services/crank.ts`: **they compose, and neither needs to wait for the other.**

Hunks are disjoint (this PR at ~38 and 1015-1022; #428 at ~853 through 1938), and `crank.ts` is the only shared file. Verified rather than eyeballed — both merged onto `main` (`052b0c5`), then `tsc --noEmit` 0, `pnpm build` 0, `pnpm test` 106 files / 1102 passed. The stall guard survives intact in the combined tree (`isStreamStalled` wired at `crank.ts:1027/1047`, `lastSlotAdvanceAt` still on `LoaderStats`), and this PR's three test files pass together there.

One number to expect if you compare: the combined tree reports 1102 tests where this PR alone reports 1143, because #428 deletes `tests/poc/m6.poc.test.ts` and `tests/services/oracle-stale.test.ts` with the dead path it removes. Not a regression from either side.

